PR #226, opened 2026-01-10
nonemptylist is stack safe

KevinBoyette:


modules/general/core/src/main/scala/oxygen/core/collection/NonEmptyList.scala
```diff
@@ -74,7 +75,40 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends PartialFunctio
...
+        case Nil    => done(acc.reverse)
+        case h :: t =>
+          tailcall(f(h)).flatMap { nel =>
+            loop(t, acc.reverse_:::(nel.toList))
+          }
+
+    loop(this.toList, Nil).map { result =>
+      NonEmptyList(result.head, result.tail)
+    }
+  }
```
Kalin-Rudnicki: Can you highlight the motivation here? 

The purpose of this class is to be a simple wrapper around `List[A]` where you know you have at least one element. A `NonEmptyList` will generally get `flatMap` called on it an average of 0-1 times, `maybe 2`, and is generally a rather small size (usually single digits).

From my understanding, the purpose of trampolining is for things like cats `Effect[_]` or zio `ZIO[_, _, _]`, so that they can essentially become your new runtime system and run forever and self recurse without blowing the stack. Thats really not the purpose of this class 😅. Effect/ZIO are like race cars, and this is like a tricycle.

Kalin-Rudnicki: With needing to to wrap `flatMapTrampoline` in this `TailRec` return type, I cant imagine a time where I would ever see normal `flatMap` and `flatMapTrampoline` (with extra things to deal with, wrapping/unwrapping `TailRec`) and say "oh, I need the beefed up version of this for this use case". Sorry, not trying to hate on the fun exercise, but this does not feel like the place for that level of optimization.

Event for the normal `flatMap` impl, Scala''s `List` has built-in mutable efficiency under the hood, and you lose that when trying to implement it yourself.

PR #293, opened 2026-08-14
OXY-155: Page switch resets scroll to top

Kalin-Rudnicki:
## OXY-155 — Page switch lands in the middle of the page

**Root cause:** `PageManager.loadPage` only scrolled after a page switch when the URL carried a `#fragment`. With no fragment it did nothing, so the new page inherited the previous page's scroll position — the reused HolyGrail center pane (`.oxy-holy-grail-center`, `overflow-y:auto`) keeps its `scrollTop`, landing the new page mid-page.

**Fix:**
- New `PageScroll` service: pure `targetFor(fragment)` → `Top | Fragment(id)`, plus `onNavigate` that either scrolls to top or to the anchor.
- New `Window.scroll.toTop` — resets the document/window scroll and the reused center scroll containers (`.oxy-holy-grail-center` / `--center-only`).
- `PageManager.loadPage` now calls `PageScroll.onNavigate(fragment)`; anchor/hash navigation behavior is unchanged.
- Added `PageScroll` pure-logic tests.

**Verification:** `sbt oxygen-ui-web/test` → 47 tests pass (incl. new suite), clean compile under `-Werror`.

**Confidence: 8/10.** High confidence on the root cause and the standard navigation path (HolyGrail + document scroll). Not higher because the JS test env has no DOM (couldn't visually confirm in a browser) and bespoke non-HolyGrail scroll containers are covered only by the window reset. See `report/OXY-155.md` for full details, decisions, and assumptions.

### DIFF
--- a/modules/ui/web/src/main/scala/oxygen/ui/web/internal/PageManager.scala
+++ b/modules/ui/web/src/main/scala/oxygen/ui/web/internal/PageManager.scala
@@ -120,8 +120,9 @@ object PageManager {
 
           _ <- pageInstance.render(navType)
 
-          // W7-T03: after first paint, scroll to #fragment (retry for async content)
-          _ <- oxygen.ui.web.service.HashScroll.toFragment(fragment).unit
+          // W7-T03 / OXY-155: after first paint, scroll to #fragment (retry for async content),
+          // or reset to the top of the page when there is no fragment (page switch).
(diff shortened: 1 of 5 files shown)


Kalin-Claude:
Superseded by #313 — branch renamed OXY-155 → context-unit/task/OFF-428 during context-framework migration (same commits).

PR #297, opened 2026-08-14
OXY-147: add JsonEncoder.Omit typeclass and @jsonOmit annotation

Kalin-Rudnicki:
## What

Adds a first-class "omit from JSON" capability to `oxygen-json`, mirroring the existing `jsonSecret`/`SecretValue`/`JsonEncoder.Secret` pattern but for **unconditional suppression** (a field that never appears on the wire) rather than redaction.

Two mechanisms:
- **`@jsonOmit` annotation** — omit a plain field without changing its type.
- **`OmitValue[A]`** opaque wrapper — the type carries omit semantics.

## Changes
- `annotations.scala`: `@jsonOmit`.
- `JsonEncoder.scala`: `def omit` + `Omit`/`OmitEncoder` (`addToObject = false`); handle `@jsonOmit` in `productDeriver`.
- `OmitValue.scala`: new opaque wrapper with encoder + decoder givens.
- `DeriveProductJsonEncoder.scala`: emit `Growable.empty` for `@jsonOmit` fields in both object-field paths (wins over `@jsonFlatten`).
- `DeriveProductJsonDecoder.scala`: omitted key dropped from `keys`; field reconstructed from `onMissingFromObject`/default.
- `JsonSchema.scala`: hide `@jsonOmit` fields from the emitted schema.
- `predef/json.scala`: export `jsonOmit` + `OmitValue`.
- Tests: `JsonSpec` omit suite + `JsonSchemaEmitterSpec` omit-hidden test.

## Key decision (needs a look)
Decode-when-key-is-present differs by mechanism:
- `@jsonOmit`: fully transient — incoming value ignored (rejected as extra under `@jsonStrict`).
- `OmitValue[A]`: encode-only — a present key is decoded normally (no annotation for the decoder macro to hook).

Omitted plain fields need a constructor default (or an `onMissing`-providing type like `Option`) to decode.

## Verification
`oxygen-jsonJVM/compile`, `oxygen-schemaJVM/compile`, `utJVM/testOnly oxygen.json.JsonSpec` (85 passed) and `oxygen.schema.*` all green.

Full write-up incl. assumptions + confidence score: `report/OXY-147.md`.

### DIFF
--- a/modules/general/json/src/main/scala/oxygen/json/JsonEncoder.scala
+++ b/modules/general/json/src/main/scala/oxygen/json/JsonEncoder.scala
@@ -61,6 +61,8 @@ trait JsonEncoder[A] {
 
   def secret: JsonEncoder.Secret[A] = JsonEncoder.Secret.fromJsonEncoder(this)
 
+  def omit: JsonEncoder.Omit[A] = JsonEncoder.Omit.fromJsonEncoder(this)
+
 }
 object JsonEncoder extends Derivable[JsonEncoder.ObjectEncoder], JsonEncoderLowPriority.LowPriority1 {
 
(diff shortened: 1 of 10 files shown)


Kalin-Claude:
Superseded by #311 — branch renamed OXY-147 → context-unit/task/OFF-297 during context-framework migration (same commits).

PR #298, opened 2026-08-14
OXY-6: array input + `= ANY(?)`

Kalin-Rudnicki:
## OXY-6 — Add support for array input + unnest

Adds first-class support in `oxygen-sql` for binding a collection as a **single** parameter and expanding it as Postgres `col = ANY(?)`, avoiding N-placeholder `IN` lists (JDBC ~32767-param limit / plan-cache blowup).

### DSL
```scala
@compile
val selectByIdArray: QueryIO[ArraySeq[UUID], Person] =
  for {
    ids <- input.array[UUID]
    p   <- select[Person]
    _   <- where if ids.contains(p.id)   // -> WHERE p.id = ANY(?)
  } yield p
```

`Q.input.array[A]` binds an `ArraySeq[A]` as one `java.sql.Array` param; `ids.contains(col)` generates `col = ANY(?)`. Encoding reuses the existing `RowRepr.ArrayRepr` / `ArraySeqEncoder` (no `::type[]` cast — the JDBC typed array already carries the element type).

### What changed
- DSL: `T.ArrayInput`, `Q.input.array`
- parsing/model: `VariableReference.ArrayFromInput`, `InputPart`, `RawQueryExpr`/`QueryExpr` `ArrayContains`, `DecoderBuilder` exhaustiveness
- generation: `FragmentBuilder` emits `<col> = ANY(?)`; `TypeclassExpr.RowRepr.arraySeq`
- tests: 3 compile-checked queries + 1 Postgres integration test (empty / 1000+ large / composed-with-scalar) — passes against a real container
- docs: array-input section in `docs/sql/queries.md`

### Scope / decisions
- Single-column element types only (per ticket "restrict to single-column A initially").
- `= ANY(?)` form only; UNNEST-as-a-table-source deferred (overlaps OXY-98 lateral join).
- Kept distinct from OXY-17 (`IN` with N placeholders).
- Full rationale + assumptions + **confidence score (8.5/10)** in `report/OXY-6.md`.

### DIFF
--- a/docs/docs/sql/queries.md
+++ b/docs/docs/sql/queries.md
@@ -118,6 +118,8 @@ input makes it a `QueryO`/`Query`. Pass `debug = true` (`@compile(debug = true)`
 | Form | Purpose |
 |------|---------|
 | `input[I]` / `input.optional[I]` / `input.const(i)` | bind a runtime / optional / compile-time-constant parameter |
+| `input.array[I]` / `input.set[I]` + `ids.contains(col)` | bind a whole `Seq` / `Set` as one array parameter, expanded to `col = ANY(?)` |
+| `select.unnest(ids)` | use a collection input as a `UNNEST(?)` join/table source (one row per element) |
 | `select[A]` | select all columns of table `A` |
 | `join[A] if <cond>` / `leftJoin[A] if <cond>` | inner / left join (`leftJoin` yields `Option[A]`) |
 | `where if <cond>` | filter |
@@ -142,6 +144,75 @@ val personJoinNotes: QueryIO[UUID, (Person, Note)] =
 
 > Custom column types flow through automatically: `input[Email]` and `select[UserRow]` use the
 > `RowRepr`/encoder/decoder for `Email` you defined in [Models](models.md).
+
+### Array input (`= ANY(?)`)
+
+To filter against a collection of values, bind the whole collection as a **single** array parameter
+with `input.array[I]` (a `Seq[I]`) or `input.set[I]` (a `Set[I]`) and test membership with
+`ids.contains(col)`. This generates Postgres `col = ANY(?)` — one bind parameter carrying a
+`java.sql.Array`, instead of expanding an `IN (…)` list to N placeholders (which hits the JDBC
+~32767-param limit and blows up the plan cache).
+
+```scala
+@compile
+val selectByIdArray: QueryIO[Seq[UUID], Person] =
+  for {
+    ids <- input.array[UUID]
+    p   <- select[Person]
+    _   <- where if ids.contains(p.id)
+  } yield p
+```
+
+```scala
+selectByIdArray.execute(Seq(id1, id2, id3))
+```
+
+Notes:
+
+- `input.array[I]` takes a `Seq[I]`, `input.set[I]` takes a `Set[I]`; either is bound as one JDBC
+  array, so callers pass their collection directly (no `ArraySeq.from(…)` conversion needed).
+- An empty collection yields `col = ANY('{}')`, i.e. no matches.
+- `I` must be a single-column type (e.g. `UUID`, `Long`, `String`, or a `RowRepr` newtype over one).
+  Composite/multi-column element types are not supported yet.
+- Array inputs compose with other inputs, e.g. `where if ids.contains(p.id) && p.groupId == groupId`.
+- No explicit `::type[]` cast is emitted: the JDBC driver builds a typed array via
+  `createArrayOf(<base type>, …)`, so `= ANY(?)` already knows the element type.
+
+### Array input as a join table (`UNNEST(?)`)
+
+The primary way to use a collection input is as a **table source** you can join against: `select.unnest(ids)`
+turns an `input.array[I]` / `input.set[I]` collection into a `UNNEST(?)` from-item that yields one row
+per element, aliased so it can be joined/filtered/selected like any other table. The whole collection
+still binds as a single `java.sql.Array` parameter.
+
+```scala
+@compile
+val notesByPersonIdUnnest: QueryIO[Seq[UUID], Note] =
+  for {
+    ids <- input.array[UUID]
+    id  <- select.unnest(ids)          // FROM UNNEST(?::uuid[]) id(id)
+    n   <- join[Note] if n.personId == id
+  } yield n
+```
+
+generates (roughly):
+
+```sql
+SELECT n.id, n.person_id, n.note
+    FROM UNNEST(?::uuid[]) id(id)
+    JOIN note n ON n.person_id = id.id
+```
+
+Notes:
+
+- The unnested element is a normal query variable — reference it in the join/where (`n.personId == id`)
+  or select it (`yield (id, n)`). It is emitted as a named, qualified column (`id.id`) so it never
+  collides with a same-named column of a joined table.
+- An empty array produces zero rows (the join matches nothing).
+- `I` must be a single-column type, same restriction as `input.array` / `input.set`.
+- Composes with other inputs/joins/wheres, e.g. an extra `input[String]` used in a `where`.
+- Currently supported as the root `FROM` source; `JOIN UNNEST(?)` (unnest as a non-root join item)
+  is not yet supported.
 
 ## A real repo method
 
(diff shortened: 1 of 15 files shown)


Kalin-Rudnicki:
## 💬 Feedback from Kalin

The **primary intended use case** for the array/unnest input — in my head — is using `UNNEST` as an **input JOIN table**, e.g.:

```sql
SELECT *
  FROM UNNEST(?) a
  JOIN other b ON b.id = a
```

This PR implemented `= ANY(?)` and **deferred** the `UNNEST`-as-table-source case — but that table-source case is the main goal, not a nice-to-have. Please make that case work and give it **solid tests** (empty array, the join actually filtering correctly, composition with other inputs/joins, real-Postgres integration).

Spinning up an agent to build the `FROM UNNEST(?) a JOIN ...` support and test it thoroughly.

Kalin-Rudnicki:
## UNNEST(?) as an input join table source (the primary use case)

Following the feedback that the main goal was using **UNNEST as an input join table** (`SELECT * FROM UNNEST(?) a JOIN other b ON b.id = a`), this is now implemented (additive — the `= ANY(?)` `contains` support is unchanged and still passing).

### DSL

`select.unnest(ids)` turns a previously-declared `input.array[A]` collection into a `UNNEST(?)` from-item that yields one row per element, aliased so it can be JOINed / filtered / selected like any other table:

```scala
@compile
val notesByPersonIdUnnest: QueryIO[ArraySeq[UUID], Note] =
  for {
    ids <- input.array[UUID]
    id  <- select.unnest(ids)          // FROM UNNEST(?::uuid[]) id(id)
    n   <- join[Note] if n.personId == id
  } yield n
```

generates:

```sql
SELECT n.id, n.person_id, n.note
    FROM UNNEST(?::uuid[]) id(id)
    JOIN note n ON n.person_id = id.id
```

The whole collection still binds as a **single** `java.sql.Array` param via the existing `RowRepr.ArrayRepr` / `ArraySeqEncoder`.

### Notes / decisions

- **Ambiguity fix:** the unnested column is named = alias and referenced qualified as `alias.alias` (via `UNNEST(?::type[]) alias(alias)`), so it never collides with a same-named column of a joined table (e.g. `note.id`).
- Keeps the array **declaration** (`input.array`) separate from its **table-source usage** (`select.unnest(ids)`) — this reuses all the array-encoder wiring and composes cleanly with other inputs/joins/wheres.
- **Single-column** element types only (v1), same restriction as `input.array`.
- Supported as the **root FROM** source (the owner's example); `JOIN UNNEST(?)` (unnest as a non-root join item) is left as future work.

### Test coverage

New integration test **"UNNEST(?) as an input join table"** in `CustomQuerySpec`, run against the real Postgres testcontainer:

- basic join filters correctly (only matching rows return)
- the unnested column decodes + is selectable (`yield (id, n)`)
- empty array → zero rows
- larger array (all real ids + 1000 random misses), still one bound param
- composition with an extra scalar `input[String]` used in a `WHERE`

**14/14 `CustomQuerySpec` tests pass** (new UNNEST test + the pre-existing `= ANY(?)` test).

docs/docs/sql/queries.md
```diff
@@ -143,6 +145,73 @@ val personJoinNotes: QueryIO[UUID, (Person, Note)] =
...
+### Array input as a join table (`UNNEST(?)`)
+
+The primary way to use an array input is as a **table source** you can join against: `select.unnest(ids)`
+turns an `input.array[I]` collection into a `UNNEST(?)` from-item that yields one row per element,
+aliased so it can be joined/filtered/selected like any other table. The whole collection still binds
+as a single `java.sql.Array` parameter.
+
+```scala
+@compile
+val notesByPersonIdUnnest: QueryIO[ArraySeq[UUID], Note] =
```
Kalin-Rudnicki: lets relax this a bit to `Seq`, doesnt it need to get transformed to an sql array anyway? no point in making people transform twice here

modules/sql/core/src/main/scala/oxygen/sql/generic/generation/FragmentBuilder.scala
```diff
@@ -306,6 +324,22 @@ final case class FragmentBuilder(inputs: List[InputPart])(using Quotes) {
...
+  def select(s: SelectPart.FromUnnest)(using ParseContext, GenerationContext, Quotes): ParseResult[GeneratedFragment] =
+    for {
+      // bind the referenced array input as a single `?` param via `RowRepr[ArraySeq[A]]` (`ArraySeqEncoder`)
+      inputFrag <- GenerationContext.updated(input = GenerationContext.Parens.Never) {
+        queryExprToFragment.input(s.arrInput, s.elemRowRepr.arraySeq)
+      }
+      // cast `?::<baseType>[]` so postgres knows the array/element type of the table source, then
+      // `<alias>(<alias>)` names the single output column so it can be referenced as `<alias>.<alias>`.
+      alias = s.mapQueryRef.sqlString
+      castAndAlias: Expr[String] = '{ "::" + ${ s.elemRowRepr.expr }.columns.columns.head.columnType.baseType + "[]) " + ${ Expr(s"$alias($alias)") } }
```
Kalin-Rudnicki: nit, put the closing paren in the `GeneratedFragment.of`

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
```diff
@@ -15,6 +16,9 @@ object Q {
 
     def const[I](i: I): T.ConstInput[I] = macroOnly
 
+    /** An array/collection input bound as a single `?` param. Use `arr.contains(col)` for `col = ANY(?)`. */
+    def array[I]: T.ArrayInput[I] = macroOnly
```
Kalin-Rudnicki, line 19: can we please have a `def set[T]` as well, relat `def array[T]` to `Seq`

report/OXY-6.md
```diff
@@ -0,0 +1,166 @@
+# OXY-6 — Add support for array input + unnest
```
Kalin-Rudnicki: move this to a comment on the JIRA issue, and remove this file

docs/docs/sql/queries.md
Kalin-Rudnicki: Done in `0f8a2209`. `input.array[I]` now binds a `Seq[I]` instead of `ArraySeq[I]` — callers pass their collection directly, and it's adapted to a JDBC array at bind time via `RowRepr.collectionAsArray` (`ArrayRepr` + `transform`), so no double-transform.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/generic/generation/FragmentBuilder.scala
Kalin-Rudnicki: Done in `0f8a2209`. The `UNNEST(` closing paren is now a literal in the `GeneratedFragment.of` (`s") $alias($alias)"`) instead of being baked into the cast string.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
Kalin-Rudnicki: Done in `0f8a2209`. Added `Q.input.set[I]` (binds a `Set[I]`) alongside `Q.input.array[I]` (now a `Seq[I]`). Both go through the same `= ANY(?)` / `UNNEST(?)` machinery — `ArrayFromInput` carries an `ArrayInputKind` (Seq | Set) and `collectionAsArray(kind)` picks the wrapper. Tests cover both spellings.

---
:robot: claude

report/OXY-6.md
Kalin-Rudnicki: Done in `0f8a2209`. Moved the report to a comment on the OXY-6 Jira issue (https://kr-oxygen.atlassian.net/browse/OXY-6) and removed `report/OXY-6.md` from the PR.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/generic/model/TypeclassExpr.scala
```diff
@@ -54,6 +54,42 @@ object TypeclassExpr {
...
+                }
+              case VariableReference.ArrayInputKind.Set =>
+                '{
+                  S.RowRepr
+                    .ArrayRepr[a]($e)
+                    .transform[scala.collection.immutable.Set[a]](
+                      (as: scala.collection.immutable.ArraySeq[a]) => as.toSet,
+                      (s: scala.collection.immutable.Set[a]) => scala.collection.immutable.ArraySeq.untagged.from(s),
+                    )
+                }
```
Kalin-Rudnicki: this really feels like it should be a non-macro function on `RowRepr`...

PR #307, opened 2026-08-15
OXY-166: Add scalar aggregate functions (SUM / AVG / MIN / MAX) to oxygen-sql

Kalin-Rudnicki:
## OXY-166 — Scalar aggregate functions (SUM / AVG / MIN / MAX)

Adds `SUM` / `AVG` / `MIN` / `MAX` scalar aggregate support to the `oxygen-sql` query DSL, mirroring the existing `COUNT` path end-to-end (DSL → parse → fragment → decode). Scalar aggregation over the whole result set — **no `GROUP BY`** (that's OXY-100).

### DSL
```scala
yield sum(p.age)   // Option[Long]
yield avg(p.age)   // Option[BigDecimal]
yield min(p.age)   // Option[Int]
yield max(p.age)   // Option[Int]
```

### Nullability
`SUM`/`AVG`/`MIN`/`MAX` return SQL `NULL` over an empty result set, so all four decode to `Option[_]` (unlike `COUNT`, which is never null).

### Type choices (documented in `docs/sql/queries.md`)
Follows Postgres' own widening:

| Aggregate | Column | PG type | Scala result |
|-----------|--------|---------|--------------|
| `sum` | Short/Int | bigint | `Option[Long]` |
| `sum` | Long/BigDecimal | numeric | `Option[BigDecimal]` |
| `sum` | Float | real | `Option[Float]` |
| `sum` | Double | double precision | `Option[Double]` |
| `avg` | Short/Int/Long/BigDecimal | numeric | `Option[BigDecimal]` |
| `avg` | Float/Double | double precision | `Option[Double]` |
| `min`/`max` | any orderable `A` | same as `A` | `Option[A]` |

The `sum`/`avg` widening is encoded by new `SumType`/`AvgType` type-classes, so the query's **static** type already reflects the widened result. The macro reads the widened `Out` type off the term and picks the matching optional decoder — the primitive `int` decoder matches on JDBC runtime class (`java.lang.Integer`), so decoding `SUM(int)` (a `bigint`/`Long`) without widening would fail; widening is required for correctness, not cosmetic.

Added a decode-only `ResultDecoder.bigDecimal` (Postgres `numeric` → `java.math.BigDecimal`). Intentionally **no** `RowRepr[BigDecimal]` / `Column.Type.Numeric` — that would ripple into the migration codecs; a `BigDecimal` table column is therefore still unsupported, but aggregate results decode to `BigDecimal` correctly.

### Tests
- New `CustomQuerySpec` "scalar aggregates" test (real Postgres via TestContainers): SUM/AVG/MIN/MAX over a non-empty group **and** an empty group (NULL → `None`), plus SUM widening (→`Long`) and AVG type (→`BigDecimal`).
- Full `sql-it` suite: **55 passed / 0 failed**.

See `report/OXY-166.md` for the full decision log and confidence score (9/10).

https://claude.ai/code/session_011YxWKdsz97QT9BD7AdpSq6

### DIFF
--- a/docs/docs/sql/queries.md
+++ b/docs/docs/sql/queries.md
@@ -128,8 +128,48 @@ input makes it a `QueryO`/`Query`. Pass `debug = true` (`@compile(debug = true)`
 | `orderBy(a.field.asc, …)`, `limit(n)`, `offset(n)` | ordering / paging |
 | `Q.insert[A]` / `Q.update[A]` / `Q.delete[A]` | begin an insert / update / delete |
 | `set(_.field := value)` | assignment in an update |
-| `count.*` / `count(a.field)` | aggregate |
+| `count.*` / `count(a.field)` | `COUNT` aggregate (result `Long`, never null) |
+| `sum(a.field)` / `sum.orNull(a.field)` / `avg(a.field)` / `min(a.field)` / `max(a.field)` | scalar aggregate |
 | `a.tablePK` / `a.tableNPK` | the row's PK / non-PK columns |
+
+### Scalar aggregates
+
+`sum` / `avg` / `min` / `max` aggregate over the **whole** result set (there is no `GROUP BY` yet).
+
+`avg` / `min` / `max` are SQL `NULL` over an empty result set, so they decode to an `Option`. `sum`
+comes in two forms:
+
+- `sum(col)` / `sum.orZero(col)` → `COALESCE(SUM(col), 0)` — a **non-null** result (`0` over an empty set).
+- `sum.orNull(col)` → `SUM(col)` — an `Option` (`None` over an empty set).
+
+```scala
+@compile
+val totalAgeInGroup: QueryIO[UUID, Long] =
+  for {
+    groupId <- input[UUID]
+    p       <- select[Person]
+    _       <- where if p.groupId == groupId
+  } yield sum(p.age)          // COALESCE(SUM(p.age), 0); 0 when the group is empty
+
+// use `sum.orNull(p.age)` for `Option[Long]` (None when the group is empty)
+```
+
+The widened `Out` type follows Postgres' own widening rules (`avg` / `min` / `max` and `sum.orNull`
+wrap it in `Option`):
+
+| Aggregate | Column type | Postgres type | Widened `Out` |
+|-----------|-------------|---------------|---------------|
+| `sum` | `Short` / `Int` | `bigint` | `Long` |
+| `sum` | `Long` / `BigInt` | `numeric` | `BigInt` |
+| `sum` | `BigDecimal` | `numeric` | `BigDecimal` |
+| `sum` | `Float` | `real` | `Float` |
+| `sum` | `Double` | `double precision` | `Double` |
+| `avg` | `Short` / `Int` / `Long` / `BigInt` / `BigDecimal` | `numeric` | `BigDecimal` |
+| `avg` | `Float` / `Double` | `double precision` | `Double` |
+| `min` / `max` | any orderable column `A` | same as `A` | `A` |
+
+The `sum` / `avg` widening is driven by the `SumType` / `AvgType` type-classes, so the query's static
+type already reflects the widened result (e.g. `sum` over an `Int` column is `Long`, `sum.orNull` is `Option[Long]`).
 
 A join example returning a tuple:
 
--- a/modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
+++ b/modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
@@ -75,6 +75,24 @@ object Q {
     def _1: Long = macroOnly
   }
 
+  // scalar aggregates over the whole result (no GROUP BY).
+
+  /**
+    * `SUM` over the whole result.
+    *   - `sum(col)` / `sum.orZero(col)` -> `COALESCE(SUM(col), 0)`, a non-null `ev.Out` (`0` over an empty result set).
+    *   - `sum.orNull(col)`              -> `SUM(col)`, `Option[ev.Out]` (SQL `NULL` -> `None` over an empty result set).
+    */
+  object sum {
+    def apply[A](toSum: A)(using ev: SumType[A]): ev.Out = macroOnly
+    def orZero[A](toSum: A)(using ev: SumType[A]): ev.Out = macroOnly
+    def orNull[A](toSum: A)(using ev: SumType[A]): Option[ev.Out] = macroOnly
+  }
+
+  // avg/min/max return `Option`: over an empty result set the aggregate is SQL NULL -> `None`.
+  def avg[A](toAvg: A)(using ev: AvgType[A]): Option[ev.Out] = macroOnly
+  def min[A](toMin: A): Option[A] = macroOnly
+  def max[A](toMax: A): Option[A] = macroOnly
+
   extension [A](self: A) {
     def tablePK(using ev: TableRepr[A]): ev.PrimaryKeyT = ev.pk.get(self)
     def tableNPK(using ev: TableRepr[A]): ev.NonPrimaryKeyT = ev.npk.get(self)
--- a/modules/sql/core/src/main/scala/oxygen/sql/query/dsl/aggregateType.scala
+++ b/modules/sql/core/src/main/scala/oxygen/sql/query/dsl/aggregateType.scala
@@ -1,0 +1,62 @@
+package oxygen.sql.query.dsl
+
+/**
+  * Type-level widening for the `SUM(_)` aggregate, mirroring Postgres' result types.
+  *
+  * Postgres widens the result of `SUM`:
+  *   - `smallint` / `int`      -> `bigint`  (Scala `Long`)
+  *   - `bigint`                -> `numeric` (Scala `BigInt`, an integral `numeric`)
+  *   - `numeric`               -> `numeric` (Scala `BigDecimal`)
+  *   - `real`                  -> `real`    (Scala `Float`)
+  *   - `double precision`      -> `double precision` (Scala `Double`)
+  *
+  * The resulting DSL expression decodes to `Option[Out]` (SQL `NULL` over an empty set -> `None`).
+  */
+sealed trait SumType[A] {
+  type Out
+}
+object SumType {
+
+  type Aux[A, B] = SumType[A] { type Out = B }
+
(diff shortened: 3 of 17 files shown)


docs/docs/sql/queries.md
```diff
@@ -118,101 +118,61 @@ input makes it a `QueryO`/`Query`. Pass `debug = true` (`@compile(debug = true)`
...
-```scala
-selectByIdArray.execute(Seq(id1, id2, id3))
-```
+Result types follow Postgres' own widening rules:
 
-Notes:
+| Aggregate | Column type | Postgres type | Scala result |
+|-----------|-------------|---------------|--------------|
+| `sum` | `Short` / `Int` | `bigint` | `Option[Long]` |
+| `sum` | `Long` / `BigDecimal` | `numeric` | `Option[BigDecimal]` |
```
Kalin-Rudnicki: this is the only one Im not OK with: `sum(Long)` -> `BigInt`

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/aggregateType.scala
```diff
@@ -0,0 +1,57 @@
+package oxygen.sql.query.dsl
```
Kalin-Rudnicki, line 1: Im ok with multiple classes in the same file in certain cases.

This one feels valid enough, fine.

If there are multiple classes in the same file, the file should have a lower-case name. The exception to this is `MyTypeclass` + `MyTypeclassLowPriority`.

Fix this file, and add that sentiment to the claude markdown scaffold.

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
```diff
@@ -75,6 +54,25 @@ object Q {
...
+  // scalar aggregates over the whole result (no GROUP BY).
+  // all return `Option`: over an empty result set the aggregate is SQL NULL -> `None`.
+
+  object sum {
+    def apply[A, B](toSum: A)(using ev: SumType.Aux[A, B]): Option[B] = macroOnly
+  }
+
+  object avg {
+    def apply[A, B](toAvg: A)(using ev: AvgType.Aux[A, B]): Option[B] = macroOnly
+  }
```
Kalin-Rudnicki: do these still work if you do it like:

```scala
def apply[A](toSum: A)(using ev: SumType[A]): Option[ev.Out] = macroOnly
```

thats kinda the whole point of the `given` Aux type pattern...

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
```diff
@@ -75,6 +54,25 @@ object Q {
...
+
+  object min {
+    def apply[A](toMin: A): Option[A] = macroOnly
+  }
+
+  object max {
+    def apply[A](toMax: A): Option[A] = macroOnly
+  }
```
Kalin-Rudnicki: also, all 4 of these dont really feel like they deserve an `object + apply`, just do `def thing[A] ...`

modules/sql/core/src/main/scala/oxygen/sql/schema/ResultDecoder.scala
```diff
@@ -145,6 +145,15 @@ object ResultDecoder extends Derivable[ResultDecoder] {
...
   given fromRowRepr: [A: RowRepr as repr] => ResultDecoder[A] = repr.decoder
 
+  /**
+    * Decoder for a Postgres `numeric` value (returned by the JDBC driver as a [[java.math.BigDecimal]]).
+    *
+    * There is intentionally no `RowRepr[BigDecimal]` (that would require a `Column.Type` + migration
+    * support). This decoder exists purely to decode the widened result of `SUM`/`AVG` aggregates.
+    */
+  val bigDecimal: ResultDecoder[BigDecimal] =
+    ResultDecoder.SingleDecoder.simplePF { case value: java.math.BigDecimal => BigDecimal(value) }
```
Kalin-Rudnicki: since this brings `BigDecimal` into scope, lets do it right:
1. Add `BigInt` and `BigInt` in `RowRepr`
2. Add tests for `BigInt` + `BigDecimal`
3. Remove this `ResultDecoder.bigDecimal`, we will have the full `RowRepr` now

report/OXY-166.md
```diff
@@ -0,0 +1,64 @@
+# OXY-166 — Add scalar aggregate functions (SUM / AVG / MIN / MAX) to oxygen-sql
```
Kalin-Rudnicki: move this report into a comment on the JIRA issue, and then delete the report file

docs/docs/sql/queries.md
Kalin-Rudnicki: Fixed in `c7fd6c36`. `sum(Long)` now widens to `BigInt` (kept integral — summing whole numbers stays whole); `sum(BigDecimal)` stays `BigDecimal`. Docs table + `SumType` updated to match, and `sum`/`avg` now also carry `BigInt`-column instances since `BigInt` is a first-class column type as of this push.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
Kalin-Rudnicki: Done in `c7fd6c36` — `def sum[A](toSum: A)(using ev: SumType[A]): Option[ev.Out]` (same for `avg`). The path-dependent `ev.Out` works; dropped the `[A, B]` from the DSL surface. The `Aux` alias still exists internally for the macro's quote pattern.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
Kalin-Rudnicki: Done in `c7fd6c36` — all four are plain `def`s now (`def sum`/`def avg`/`def min`/`def max`), no more `object` + `apply`.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/schema/ResultDecoder.scala
Kalin-Rudnicki: Done all three in `c7fd6c36`:
1. Added `RowRepr[BigInt]` + `RowRepr[BigDecimal]`, backed by a new `Column.Type.Numeric` threaded through the migration `ColumnColumn` codec + both `domainToDb`/`dbToDomain` conversions.
2. Tests: `RowSchemaSpec` numeric column-type asserts + a real-Postgres `BigInt`/`BigDecimal` round-trip in `CustomQuerySpec` (values beyond `Long`/`Double` range to catch a lossy codec).
3. Removed `ResultDecoder.bigDecimal` — aggregate decoding now goes through the real `RowRepr`s.

---
:robot: claude

report/OXY-166.md
Kalin-Rudnicki: Done — moved the report into a comment on the OXY-166 JIRA issue (rewritten to reflect the post-review design) and deleted `report/OXY-166.md` in `c7fd6c36`.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/aggregateType.scala
Kalin-Rudnicki: Renamed `AggregateType.scala` → `aggregateType.scala` in `c7fd6c36` (lower-case, since it holds `SumType` + `AvgType`).

On the second part — I couldn't find a "claude markdown scaffold" anywhere in the repo (no `CLAUDE.md`, and nothing under `agent-docs/` covers Scala file conventions). Where does that scaffold live, or where would you like the convention added — *"multiple top-level classes in one file ⇒ lower-case file name, except `MyTypeclass` + `MyTypeclassLowPriority`"*? I'll add it once I know the target. Leaving this open for that.

---
:robot: claude

Kalin-Rudnicki: Resolved the scaffold question in `9eddef1c`: stubbed a minimal `CLAUDE.md` + `CLAUDE/scala-file-naming.md` capturing the rule (multiple top-level classes ⇒ lower-case file name, except `MyTypeclass` + `MyTypeclassLowPriority`). Kept it deliberately dead-simple; filed **OXY-168** to come back and flesh `CLAUDE/` out into a real convention guide.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/query/dsl/Q.scala
```diff
@@ -75,6 +54,14 @@ object Q {
     def _1: Long = macroOnly
   }
 
+  // scalar aggregates over the whole result (no GROUP BY).
+  // all return `Option`: over an empty result set the aggregate is SQL NULL -> `None`.
+
+  def sum[A](toSum: A)(using ev: SumType[A]): Option[ev.Out] = macroOnly
```
Kalin-Rudnicki: I changed my mind on `sum`, this one `is` worth an `object` (the others still arent).

```scala
object sum {
  def orZero[A] (): Out
  def orNull[A])(): Option[Out]
  def apply[A](): Out
}
```

`orZero` / `apply` -> `COALESCE`
`orNull` -> same as now

Kalin-Rudnicki: Done in `91783a5f` — `sum` is now an `object` (avg/min/max stay plain defs):

```scala
object sum {
  def apply[A](toSum: A)(using ev: SumType[A]): ev.Out          // COALESCE(SUM(_), 0)
  def orZero[A](toSum: A)(using ev: SumType[A]): ev.Out         // COALESCE(SUM(_), 0)
  def orNull[A](toSum: A)(using ev: SumType[A]): Option[ev.Out] // SUM(_) — as before
}
```

- `sum(col)` / `sum.orZero(col)` → `COALESCE(SUM(col), 0)`, non-optional `ev.Out` (`0` over an empty set).
- `sum.orNull(col)` → `SUM(col)`, `Option[ev.Out]` (`None` over an empty set).

Threaded a `coalesceZero` flag through the aggregate pipeline (parser → `QueryExpr` → fragment → decoder): the fragment emits `COALESCE(…, 0)` and the decoder picks a non-optional result decoder for that variant. Tests cover both `sum` (→ 0 on empty) and `sum.orNull` (→ None on empty); docs updated.

---
:robot: claude

PR #292, opened 2026-08-14
OXY-146: Pulsar createIfDNE idempotency (Topic already exists)

Kalin-Rudnicki:
## Summary
Fixes OXY-146: `PulsarAdminClient.{tenant,namespace,topic}.createIfDNE` used a non-atomic check-then-act (list then create). Under concurrent calls both fibers could see the resource missing and both create it; the loser got a 409 `PulsarAdminException` ("... already exists"). Now treated as success.

## Changes
- Shared private `succeedIfAlreadyExists` combinator wraps the `create` call in all three `createIfDNE` methods (`catchSome` -> log Info, return unit).
- Private `isAlreadyExists` predicate: `PulsarAdminException` status 409 first, unwrap `CompletionException`/`ExecutionException`, then case-insensitive "already exists" message fallback.
- Preserved behavior: partition-mismatch still warns + succeeds; list-based fast-path unchanged; non-already-exists errors still propagate.
- Added first test dir for `oxygen-events-pulsar` (+ `oxygen-test` % Test): mock-based idempotency + propagation tests via JDK dynamic proxy. 6 tests pass.

## Notes
- Decisions + build caveats documented in `report/OXY-146.md`.
- Confidence: 8.5/10.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #310, opened 2026-08-21
OFF-425: Pulsar `createIfDNE` throws random error - Topic already exists

Kalin-Claude:
Framework migration of the OXY-146 work into the context-framework scaffold.

Supersedes #292 (same commits; branch renamed OXY-146 → context-unit/task/OFF-425).

Jira: OFF-425


Kalin-Claude:
## 🤖 Standards review (scala + oxygen)

Reviewed the diff (`build.sbt`, `PulsarAdminClient.scala`, `PulsarAdminClientSpec.scala`) against `~/.claude/standards/{scala,oxygen}`. One hard-rule violation and a few nits; the `build.sbt` change and the overall shape of the change are clean.

### Hard-rule violations

- **`modules/events/pulsar/src/main/scala/oxygen/pulsar/client/PulsarAdminClient.scala:166`** — [scala/hard-rules.md §5: no `null`/`== null`/`!= null`] `error.getCause != null` uses a raw null check. This is doubly worth fixing because the very next line (169) already reaches for the idiomatic `Option(error.getMessage)` — so the two adjacent branches are inconsistent.
  - Fix: `Option(error.getCause).filter(_ ne error).exists(isAlreadyExists)` (folds the null-guard, the `ne error` self-reference guard, and the recursion into one expression, and matches line 169's style).

### Nits

- **`PulsarAdminClientSpec.scala:20-38` (the `proxy` helper)** — [oxygen/testing/index.md: "hand-write in-memory/recording doubles as classes … No mocking library is used"] The `java.lang.reflect.Proxy` + string-name dispatch is effectively a dynamic mock, and it pulls in three further carve-outs: string-dispatch on method names (`"getListAsync"`, `"topics"`, …) brushes [hard-rules §2 stringly-typed]; `proxyInstance.asInstanceOf[AnyRef]` (line 32) is [hard-rules §3 `.asInstanceOf`]; and `throw new UnsupportedOperationException` (line 34) is [hard-rules §8 no `throw`]. All three are genuinely at the Java-reflection interop edge (the documented library carve-out), and PulsarAdmin is a huge interface so the pragmatism is understandable — so this is a judgment-call observation, not a clear violation. If a small hand-written class stub of just the `Topics`/`Namespaces`/`Tenants` surface used here is feasible, it would sit better with the "class doubles, no mock lib" convention and drop the cast/throw/string-match all at once.

- **`PulsarAdminClient.scala:163`** — [scala/syntax.md: prefer braceless `match` with vertically-aligned `=>`; braces only for multi-statement arms] Every arm in `isAlreadyExists`'s `match` is a single expression, so the `{ }` around the `match` isn't needed — a braceless `error match` is the house default.

- **`PulsarAdminClient.scala:54,81,133` (+ `ZIO.logInfo(alreadyExistsMessage)`)** — [oxygen/index.md anti-patterns: use `str"…"` (Oxygen `Text`) for logged output, not `s"…"`] The new "already exists" messages are built with `s"…"`. Very low priority: the whole file already logs with `s"…"`, so this matches local convention rather than introducing a new deviation — flagging only for completeness against the standard.

Test structure itself is on-spec: `object … extends OxygenSpecDefault`, single `import oxygen.predef.test.*`, lowercase behavioral test names, `assert(exit)(succeeds/fails(anything))`, and outer suite named after the spec. Explicit return types are present throughout both files. 👍

PR #304, opened 2026-08-14
OXY-162: New Oxygen command-execution API (CommandService / Command2)

Kalin-Rudnicki:
Derives from the **`current/feature/command-improvements`** brainstorm branch (a chain of
"WIP : savepoint" commits) and brings that exploratory work to a compiling, tested state.

Jira: [OXY-162](https://kr-oxygen.atlassian.net/browse/OXY-162)

## What the new command API is

A "v2" API in `oxygen-zio` for running external OS processes from ZIO, intended to eventually
replace the legacy `oxygen.zio.system.Command` (which shells out via `scala.sys.process`):

- **`CommandService`** — pluggable execution SPI (`executeSync`, `executeSyncStreamErr`,
  `executeCode` + derived helpers), held in a `FiberRef` with a platform-specific default.
- **`Command2`** — immutable fluent builder (command / args / cwd / env / sudo) building a
  `BuiltCommand`; execute surface for sync capture, streamed stderr, exit-code, and typed
  decode-to-`A` (String / PlainText schema / JSON), plus legacy-compatible `execute*` shims.
- **`JavaCommandService`** — JVM/Native impl over `java.lang.ProcessBuilder`: scoped process
  lifecycle, concurrent stdout/stderr draining (avoids pipe-buffer deadlock), stdin sources.
- **`UnimplementedCommandService`** — Scala.js default (typed `Unimplemented` failure).
- stdio ADTs (`CommandInputSource` / `CommandOutputSource` / `ShowCommand`) and a `CommandError` ADT.
- Supporting: `StringDecoder.DecodingFailure` and `json.JsonError` now extend oxygen `Error`.

## What this PR did to finish it

- **Fixed a blocking bug:** `CommandInputSource.Empty` mapped to `Redirect.DISCARD` (a WRITE-only
  redirect), so the process failed to *start* for every no-stdin execution — i.e. the API failed on
  first use. Now uses `PIPE` + closes the child's stdin (EOF).
- **Restored JS/Native compilation:** the shared `JavaCommandService` used `ZSink.fromOutputStream`
  (JVM-only) and `ProcessBuilder` (absent on JS). Rewrote the stream-stdin path to a portable
  blocking chunk-write (Native compiles it), and excluded the file from the JS source set.
- **Resolved WIP loose ends:** corrected `BuiltCommand` shell-escaping to the POSIX `'\''` idiom
  (display/logging only — execution never goes through a shell), documented the `Command2` vs legacy
[...]

PR #312, opened 2026-08-21
OFF-314: New Oxygen command-execution API (CommandService / Command2)

Kalin-Claude:
Framework migration of the OXY-162 work into the context-framework scaffold.

Supersedes #304 (same commits; branch renamed OXY-162 → context-unit/task/OFF-314).

Jira: OFF-314

### DIFF
--- a/modules/general/zio/src/main/scala/oxygen/zio/system/JavaCommandService.scala
+++ b/modules/general/zio/src/main/scala/oxygen/zio/system/JavaCommandService.scala
@@ -1,0 +1,305 @@
+package oxygen.zio.system
+
+import java.io.{InputStream, OutputStream}
+import java.lang as jl
+import java.nio.charset.StandardCharsets
+import oxygen.core.StringBuilder
+import oxygen.predef.core.*
+import oxygen.zio.ZioCauses
+import oxygen.zio.error.CommandError
+import oxygen.zio.syntax.error.*
+import oxygen.zio.syntax.log.*
+import zio.*
+import zio.stream.*
+
+object JavaCommandService extends CommandService {
+
+  private val charset = StandardCharsets.UTF_8
+
+  //////////////////////////////////////////////////////////////////////////////////////////////////////
+  //      Public API
+  //////////////////////////////////////////////////////////////////////////////////////////////////////
+
+  override def executeSync(
+      command: BuiltCommand,
+      stdIn: CommandInputSource,
+      trim: Boolean,
+  )(using Trace): IO[CommandError, (stdOut: String, stdErr: String, exitCode: Int)] =
+    for {
+      outBuilder <- ZIO.succeed { StringBuilder.emptyThreadUnsafe }
+      errBuilder <- ZIO.succeed { StringBuilder.emptyThreadUnsafe }
+      exitCode <- run(
+        command = command,
+        stdIn = stdIn,
+        stdOut = OutputMode.Collect(outBuilder),
+        stdErr = OutputMode.Collect(errBuilder),
+      )
+    } yield (
+      stdOut = maybeTrim(outBuilder.build(), trim),
+      stdErr = maybeTrim(errBuilder.build(), trim),
+      exitCode = exitCode,
+    )
+
+  override def executeSyncStreamErr(
+      command: BuiltCommand,
+      stdIn: CommandInputSource,
+      stdErr: CommandOutputSource,
+      trim: Boolean,
+  )(using Trace): IO[CommandError, (stdOut: String, exitCode: Int)] =
+    for {
+      outBuilder <- ZIO.succeed { StringBuilder.emptyThreadUnsafe }
+      stdErrMode <- OutputMode.fromSource(command, stdErr)
+      exitCode <- run(
+        command = command,
+        stdIn = stdIn,
+        stdOut = OutputMode.Collect(outBuilder),
+        stdErr = stdErrMode,
+      )
+    } yield (
+      stdOut = maybeTrim(outBuilder.build(), trim),
+      exitCode = exitCode,
+    )
+
+  override def executeCode(
+      command: BuiltCommand,
+      stdIn: CommandInputSource,
+      stdOut: CommandOutputSource,
+      stdErr: CommandOutputSource,
+  )(using Trace): IO[CommandError, Int] =
+    for {
+      stdOutMode <- OutputMode.fromSource(command, stdOut)
+      stdErrMode <- OutputMode.fromSource(command, stdErr)
+      exitCode <- run(
+        command = command,
+        stdIn = stdIn,
+        stdOut = stdOutMode,
+        stdErr = stdErrMode,
+      )
+    } yield exitCode
+
+  //////////////////////////////////////////////////////////////////////////////////////////////////////
+  //      Core runner
+  //////////////////////////////////////////////////////////////////////////////////////////////////////
+
+  private def run(
+      command: BuiltCommand,
+      stdIn: CommandInputSource,
+      stdOut: OutputMode,
+      stdErr: OutputMode,
+  )(using Trace): IO[CommandError, Int] =
+    ZIO.scoped {
+      for {
+        inRedirect <- inputRedirect(command, stdIn)
+        outRedirect = outputRedirect(stdOut)
+        errRedirect = outputRedirect(stdErr)
+        process <- startProcess(command, inRedirect, outRedirect, errRedirect)
+
+        // Stdout/stderr must be consumed concurrently with waitFor to avoid pipe-buffer deadlocks
+        // whenever those streams are PIPE'd into our process.
+        outFiber <- consumeOutput(command, process.getInputStream, stdOut).forkScoped
+        errFiber <- consumeOutput(command, process.getErrorStream, stdErr).forkScoped
+        inFiber <- writeInput(command, process, stdIn).forkScoped
+
+        exitCode <-
+          ZIO
+            .attemptBlockingInterrupt { process.waitFor() }
+            .convertCausesFail { executionFailure(command, "wait for process exit", _) }
+
+        // Process is done — stop feeding stdin; still drain any piped stdout/stderr fully.
+        _ <- inFiber.interrupt
+        _ <- outFiber.join
+        _ <- errFiber.join
+      } yield exitCode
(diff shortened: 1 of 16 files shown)


Kalin-Claude:
🤖 **Standards review (scala + oxygen)**

Reviewed the 14-file diff (`origin/main...HEAD`, ~985 insertions — the new `CommandService`/`Command2` API + spec). No **hard-rule** violations found — the design cleanly mirrors the existing sibling `FileSystem` pattern in the same package (typed `Error` hierarchy, `FiberRef` service ref, per-platform `*PlatformSpecificImpl`, `Unimplemented*` fallback), stays braced, uses `given`/`using`, typed error channel, no `var`/`null`/`throw`/`asInstanceOf`, and the `execute*` default params are the sanctioned "lean helper forwarding to a no-default verbose method" carve-out. The two in-code `// TODO (KR) :` markers are correctly spelled and there are **no** `FIX-PRE-MERGE` markers left in the diff.

A few minor nits (all optional / judgment calls, none blocking):

**Nits**

1. `modules/general/zio/src/main/scala/oxygen/zio/system/JavaCommandService.scala:280` — [oxygen anti-patterns: `str"…"` over `s"…"` for logged/error output; errors modeling] `Error(s"Failed to $whileAttemptingTo", causes)` builds a logged error message with `s"…"`. The sibling `FileSystemError.GenericOperationError` instead carries `whileAttemptingTo`/context as *fields* and renders via `str"…"` in `errorMessage`. Consider matching that: either `str"Failed to $whileAttemptingTo"`, or model the context on a `CommandError` case rather than wrapping a generic inner `Error`.

2. `JavaCommandService.scala:210` and `:251` — [effects/resource-management: "never manual try/finally"] two `try { … } finally …close()` blocks for stream cleanup, whereas the parallel `CommandInputSource.Stream` case at `:221` already uses the ZIO idiom `.ensuring { … .ignore }`. Low-level JVM interop is a recognized carve-out (`Text.scala`/`quoted/Tree.scala` also use `finally`), so this is a consistency nit rather than a violation — the two could switch to `.ensuring` to match line 221.

3. `command.scala` (`BuiltCommand` @ :10 + `Command2` @ :59) and `stdio.scala` (`CommandInputSource`/`CommandOutputSource`/`ShowCommand` @ :9/:28/:49) — [naming-and-files: one-type-per-file] lowercase filenames legitimately signal grouping, but the guideline is "don't put several *substantive* models in one file." These hold multiple substantive types each. They're tightly coupled (`Command2.build` ↔ `BuiltCommand` + the `Conversion`; the three stdio ADTs are one config cluster), so it's defensible — flagging only as a split-candidate (`BuiltCommand.scala` / `Command2.scala`) if you want strict one-type-per-file.

4. `CommandService.scala:30` — [formatting/section-banners] the subsection banner `///////  ///////…` has an empty label; the house form carries a label (`/////// Label ///////`). Trivial.

modules/general/zio/src/main/scala/oxygen/zio/system/command.scala
```diff
@@ -0,0 +1,276 @@
...
+
+/**
+  * Immutable builder for an external OS command, executed through [[CommandService]].
+  *
+  * This is the "v2" command API: unlike the legacy [[Command]] (which shells out via `scala.sys.process`),
+  * execution goes through the pluggable [[CommandService]] with first-class stdin/stdout/stderr sources and
+  * typed decoding of process output. It is intentionally kept alongside the legacy [[Command]] until all
+  * call-sites have migrated; the `Command2` name is a deliberate interim so the two can coexist.
+  */
+final class Command2 private (isSudo: Boolean, command: String, args: Growable[String], cwdPath: Option[Path], env: Growable[(String, String)]) {
```
Kalin-Rudnicki: MUST BE RESOLVED BEFORE MERGE - can not have this called `Command2`

modules/general/zio/src/main/scala/oxygen/zio/system/JavaCommandService.scala
```diff
@@ -0,0 +1,314 @@
...
+  //////////////////////////////////////////////////////////////////////////////////////////////////////
+
+  override def executeSync(
+      command: BuiltCommand,
+      stdIn: CommandInputSource,
+      trim: Boolean,
+  )(using Trace): IO[CommandError, (stdOut: String, stdErr: String, exitCode: Int)] =
+    for {
+      outRef <- Ref.make("")
+      errRef <- Ref.make("")
```
Kalin-Rudnicki: absolutely not - way too inefficient - use oxygen `StringBuilder`

modules/general/zio/src/main/scala/oxygen/zio/system/command.scala
```diff
@@ -0,0 +1,276 @@
+package oxygen.zio.system
+
+import oxygen.json.JsonDecoder
+import oxygen.predef.core.*
+import oxygen.schema.*
+import oxygen.zio.error.CommandError
+import oxygen.zio.logging.LogLevels
+import zio.*
+
+final case class BuiltCommand(
```
Kalin-Rudnicki: split into separate file, and make sure naming follows conventions

modules/general/zio/src/main/scala/oxygen/zio/system/command.scala
```diff
@@ -0,0 +1,276 @@
...
+    * the rendered string could be safely pasted into a shell.
+    */
+  def safeShow(value: String, forceEscape: Boolean): String = {
+    val needsEscape: Boolean =
+      forceEscape || value.isEmpty || value.exists {
+        case '\'' | '"' | ' ' | '\t' | '\n' => true
+        case _                              => false
+      }
+
+    if needsEscape then s"'${value.replace("'", "'\\''")}'"
```
Kalin-Rudnicki: do we have tests for proper handling of `\\`, `\\\\`, `\\'`, `\'`, etc?

modules/general/zio/src/main/scala/oxygen/zio/system/JavaCommandService.scala
```diff
@@ -0,0 +1,314 @@
...
+          .runCollect
+          .map { chunk => new String(chunk.toArray, charset) }
+          .flatMap(ref.set)
+          .convertCausesFail { executionFailure(command, "read process output", _) }
+
+      case OutputMode.PipeTo(target) =>
+        ZIO
+          .attemptBlockingInterrupt {
+            try stream.transferTo(target)
+            finally stream.close()
```
Kalin-Rudnicki: naughty naught `try`

modules/general/zio/src/main/scala/oxygen/zio/system/command.scala
Kalin-Claude: Renamed `Command2` → `Command` in `2a13099c`. The old `scala.sys.process` runner became `LegacyCommand` — the `oxygen.predef.zio` export is now `{Command, LegacyCommand, Path}`, and its remaining call-sites (`AutoComplete`, `TestContainerService`, `PostgresTestContainer`) point at `LegacyCommand`. Kept the legacy impl under the new name rather than deleting it, so migrating those call-sites onto the new API stays a separate step.

---
:robot: claude

modules/general/zio/src/main/scala/oxygen/zio/system/JavaCommandService.scala
Kalin-Claude: Done in `2a13099c`. `OutputMode.Collect` now accumulates into an oxygen `StringBuilder` (`emptyThreadUnsafe` — single writer fiber, read after `.join`) fed by `ZPipeline.utf8Decode` + `runForeach`, instead of `Ref[String]` + `runCollect` materializing the whole byte `Chunk` and copying it into a `String`.

---
:robot: claude

modules/general/zio/src/main/scala/oxygen/zio/system/command.scala
Kalin-Claude: Split in `2a13099c`: `command.scala` → `BuiltCommand.scala` + `Command.scala` (one top-level type per file, PascalCase filenames). The legacy runner moved to `LegacyCommand.scala`, and the specs follow the same convention (`CommandSpec` for the new API, `LegacyCommandSpec` for the old).

---
:robot: claude

modules/general/zio/src/main/scala/oxygen/zio/system/command.scala
Kalin-Claude: Good catch — there were none, and the deny-list was actually **wrong** for backslashes: a bare `a\b` was returned un-quoted, so pasting it into a shell mangles it to `ab`. Fixed in `2a13099c` by flipping `safeShow` to an allow-list (alphanumerics + `@%+=:,./-_`, matching `shlex.quote`); everything else — `\`, `$`, backticks, globs, redirections, whitespace — now forces single-quoting. Added `CommandSpec` cases for `a\b`, `a\\b`, `a\'b`, `$HOME`, `a*b`, backticks, plus a safe-punctuation case that must stay bare.

---
:robot: claude

modules/general/zio/src/main/scala/oxygen/zio/system/JavaCommandService.scala
Kalin-Claude: Removed in `2a13099c` — this one and its sibling `try`/`finally` in `writeInput`'s `Const` case are now `.ensuring { ZIO.attempt { …close() }.ignore }`.

---
:robot: claude

modules/general/zio/src/main/scala/oxygen/zio/system/command.scala
Kalin-Claude: Follow-up in `8cbcf1e9`: went the whole way and removed `LegacyCommand` outright. The three remaining call-sites — `chmod` (AutoComplete), `docker` (TestContainerService), `psql` (PostgresTestContainer) — now use the new `Command`, and it's dropped from the `oxygen.predef.zio` export (back to `{Command, Path}`).

No call-site signatures changed: `CommandError <: Error <: Throwable`, so `IO[CommandError, _]` is a subtype of `Task[_]` — the new `execute*` API is a strict narrowing of the old `Throwable` channel. Compiles on JVM/JS/Native + downstream, tests green.

---
:robot: claude

PR #313, opened 2026-08-21
OFF-428: Page switch lands in the middle of the page

Kalin-Claude:
Framework migration of the OXY-155 work into the context-framework scaffold.

Supersedes #293 (same commits; branch renamed OXY-155 → context-unit/task/OFF-428).

Jira: OFF-428


Kalin-Claude:
🤖 **Standards review (scala + oxygen)**

Reviewed the diff vs `origin/main` (4 files: `PageManager.scala`, `PageScroll.scala`, `Window.scala`, `OxygenColorSystemSpec.scala`). No hard-rule violations. A few style/consistency nits below.

**Nits**

- `modules/ui/web/src/main/scala/oxygen/ui/web/service/PageScroll.scala:22` and `:29` — [scala/syntax.md — `match`] both matches use a braced `match { … }`, but every arm is a single-line expression and the match is the direct RHS of a `def`. House style is **braceless** `match` with aligned `=>`; braces are only for multi-statement arms or when the match is an argument. → drop the `{ }` on both `targetFor` and `onNavigate`.

- `modules/ui/web/src/main/scala/oxygen/ui/web/service/Window.scala:96-104` (`resetScrollContainers`) — [scala/hard-rules.md #6 — no `var`/mutable state] `var i = 0` + `while i < nodes.length do { … i += 1 }`. This is Scala.js DOM interop, which is a legitimate library carve-out, and mutating `el.scrollTop/scrollLeft` is unavoidable — but the *loop counter* `var` is not. The rest of this package iterates with `.foreach` (e.g. line 119); this is the only `var`/`while` in `ui/web/service`. → prefer `(0 until nodes.length).foreach { i => val el = nodes(i); el.scrollTop = 0; el.scrollLeft = 0 }` (safe path + consistent with local idiom).

- `modules/ui/web/src/main/scala/oxygen/ui/web/service/Window.scala:142` (`toTop`) — [scala/hard-rules.md #1 — default parameters] `smooth: Boolean = false` is a default arg. Borderline-acceptable here: it's a builder-style animation toggle and mirrors the sibling `toY(…, smooth: Boolean = true)`/`toX` already in this object, so it's consistent local convention rather than the silent-slip trap. Noting for awareness only — no change required unless you want to tighten the whole family.

**Looks good**

- `PageScroll` — one-type-per-file, braced `object`/`enum` bodies, explicit return types on every `def`, SpaceAsterisk scaladoc with `[[Target.Top]]` refs, pure decision (`targetFor`) split from the effectful `onNavigate`, no `derives` on the pure internal `Target` enum. All on-standard.
- Test suite uses `assertTrue` with typed `PageScroll.Target` comparisons — no stringly-typed assertions, good coverage of the trim/`#`/blank edge cases.

PR #316, opened 2026-08-21
OFF-325: Add compatibility checking to db migrations

Kalin-Claude:
## What & why

OFF-325 asks for compatibility checking on db migrations (add / remove / alter columns → schema compatibility). Grounding found the feature is **already fully implemented and merged on `main`**: `MigrationGenerator.classify`/`bump`, the `MigrationCompatibility` enum, `MigrationCheck` with `OXYGEN_MIGRATION_ALLOW_UPDATE` / `OXYGEN_MIGRATION_ALLOW_INCOMPATIBLE` gating, and `PersistedMigrationFile.compatibility`. So the remaining work is **verification/closure**, not green-field.

## What changed

Adds verification-hardening unit tests to `MigrationGeneratorSpec` that drive `MigrationGenerator.classify` directly (pure, no DB) over hand-built `StateDiff`s, covering the classification-table branches the existing spec did not exercise:

- the same-migration **`createdTables` exception** — FK / unique index are `BackwardsCompatible` only on a table created in the same migration, else `Incompatible`;
- `SetNullable` relax (compatible) vs tighten (incompatible);
- `CreateColumn` nullable vs `NOT NULL`;
- drops: column/table/schema `Incompatible`, foreign-key/index `BackwardsCompatible`;
- `RenameColumn` incompatible; non-unique index always compatible;
- the aggregate rule: any single incompatible diff ⇒ the whole migration is `Incompatible`.

No production code is touched — this turns the "verify the classification table matches the intended contract" ask into executable assertions + a regression guard. No behavior change ⇒ no doc / migration-json edits.

**Deferred** (agent-proposed design decisions, left for a human call — not in scope for this 1-pt task): default-value-aware `NOT NULL` classification (`Column` has no default concept yet), apply-time gating in `MigrationService`, and a richer multi-level compat lattice.

## Verification

`sbt "scalafmtAll; sql-it/testOnly oxygen.sql.migration.MigrationGeneratorSpec"` → compiles clean, **12 passed / 0 failed** (4 original + 8 new).

Jira: OFF-325

### DIFF
--- a/modules/sql/it-test/src/test/scala/oxygen/sql/migration/MigrationGeneratorSpec.scala
+++ b/modules/sql/it-test/src/test/scala/oxygen/sql/migration/MigrationGeneratorSpec.scala
@@ -3,7 +3,7 @@ package oxygen.sql.migration
 import oxygen.core.Version
 import oxygen.predef.test.*
 import oxygen.sql.migration.MigrationGenerator.GenerateResult
-import oxygen.sql.migration.model.MigrationState
+import oxygen.sql.migration.model.*
 import oxygen.sql.migration.persistence.model.{MigrationCompatibility, MigrationStepColumn}
 import oxygen.sql.query.TableCompanion
 import oxygen.sql.schema.*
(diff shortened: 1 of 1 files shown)


Kalin-Claude (commented):
Reviewer pass (adversarial, standards-focused). Test-only change: 8 new pure unit tests driving `MigrationGenerator.classify` over hand-built `StateDiff`s.

**Verified — no findings:**
- Every one of the 8 assertions matches the production `classifyOne` table exactly (`MigrationGenerator.scala:64-92`): CreateColumn nullable/NOT-NULL, SetNullable relax/tighten, the column/table/schema drops (Incompatible) vs FK/index drops (BackwardsCompatible), RenameColumn, the `createdTables` same-migration exception for FK + unique index, non-unique index always compatible, and the aggregate any-incompatible rule. All correct.
- Fixtures are constructor-correct against the models: `Column(name, Type.Int, nullable)`, `TableRef(schema, table)` (via the `apply(schemaName, tableName)` overload), `ColumnRef(table, name)`, `ForeignKeyState(None, self, references, pairs)`, `IndexState(None, self, unique, cols)`, and `.ref` on FK/index. `fkOn(personRef).self == personRef == CreateTable(personTable).tableName`, so the same-migration cases resolve as intended.
- No hard-rule violations in the added lines: explicit return types on every helper, wildcard import is standards-preferred, no `throw`/`null`/`var`/default-params/`asInstanceOf`/stringly-typed matching introduced. (The two `throw new RuntimeException` lines are pre-existing context, not part of this diff.)

One non-blocking coverage note inline.

---
:robot: claude

modules/sql/it-test/src/test/scala/oxygen/sql/migration/MigrationGeneratorSpec.scala
```diff
@@ -70,6 +82,60 @@ object MigrationGeneratorSpec extends OxygenSpecDefault {
           g.file.previousVersion.contains("1.0.0"),
         )
       },
+      suite("classify (classification table)")(
```
Kalin-Claude, line 85: **minor** — The PR frames this as verifying "the classification table", but 8 of the 18 `classifyOne` branches remain unexercised after this suite (and the 4 pre-existing tests): the additive `AlterExtension.CreateExtension` and `AlterSchema.CreateSchema` (→ BackwardsCompatible), and all six rename branches — `AlterForeignKey.Rename{Explicitly,Auto}NamedForeignKey`, `AlterIndex.Rename{Explicitly,Auto}NamedIndex` (→ BackwardsCompatible), and the defensively-Incompatible `AlterSchema.RenameSchema` / `AlterTable.RenameTable` (`MigrationGenerator.scala:68-91`). The two defensive rename→Incompatible branches are the most valuable to pin down, since a future refactor could silently flip them to compatible with no test catching it. Not blocking — the covered branches are the ones that actually appear in auto-generated diffs — but worth a line if the intent is a full regression guard on the table.

---
:robot: claude

PR #317, opened 2026-08-21
OFF-320: Implement prepared statement caching

Kalin-Claude:
Adds a typed, opt-in surface for pgjdbc **server-side prepared-statement caching** on `DbConfig.Connection`.

## What & why
pgjdbc already keeps a per-connection cache of server-prepared statements keyed by SQL text (it survives the client-side `PreparedStatement.close()`) and promotes a statement to a server-prepared plan after `prepareThreshold` executions. Because oxygen-sql pools and reuses connections, the Postgres `Parse` cost is already amortized at pgjdbc's defaults — what was missing is a **typed, documented way to tune or disable it**. This PR adds exactly that; it is not caching-from-scratch.

**Driver-level only.** No app-level per-connection LRU: it would duplicate pgjdbc's (transaction-safe) cache and drag in the `Scope`-cleanup and `Atomically` rollback/savepoint invalidation hazards the issue flags, for no benefit.

## Changes
- `DbConfig.PreparedStatementCache(prepareThreshold, cacheQueries, cacheSizeMiB: Option[Int])` — derives `JsonSchema`, `.properties` → `prepareThreshold` / `preparedStatementCacheQueries` / `preparedStatementCacheSizeMiB`.
- Nested `preparedStatementCache` field on `DbConfig.Connection`; its pairs are appended in `Connection.properties` **before** `extraProperties`, so the escape hatch still wins on collision. `Driver.buildProperties` funnels these unchanged.
- Every field is optional; **omitting the block is a no-op** that preserves pgjdbc defaults. Set the cache fields to `0` to disable client-side caching.
- Docs: `docs/docs/sql/database.md` gains a knob/property/default table + JSON example.
- Tests: `DbConfigSpec` gains a `PreparedStatementCache` suite — translation, disable-via-zero, `extraProperties` precedence, `buildProperties` threading, JSON decode. Full spec is green (18 tests).

## Open questions from the ticket, resolved
1. Driver-only (not app-level LRU). 2. Opt-in, preserving pgjdbc defaults. 3. Typed config (generic passthrough already exists as `extraProperties`). 4. N/A — pgjdbc keys internally. 5. N/A — pgjdbc manages server-statement lifecycle across tx/savepoint.

Jira: OFF-320

### DIFF
--- a/modules/sql/core/src/main/scala/oxygen/sql/DbConfig.scala
+++ b/modules/sql/core/src/main/scala/oxygen/sql/DbConfig.scala
@@ -45,6 +45,7 @@ object DbConfig {
     * @param connectTimeout Max time to establish the TCP connection (`connectTimeout`, whole seconds).
     * @param socketTimeout  Max time a socket read may block (`socketTimeout`, whole seconds).
     * @param applicationName Reported to the server as `ApplicationName` (shows up in `pg_stat_activity`).
+    * @param preparedStatementCache pgjdbc server-side prepared-statement caching knobs (see [[PreparedStatementCache]]).
     * @param extraProperties Escape hatch: arbitrary driver properties copied verbatim. Applied last, so these
     *                        override the typed properties (and credentials) on key collision.
     */
@@ -56,6 +57,7 @@ object DbConfig {
       connectTimeout: Option[Duration] = None,
       socketTimeout: Option[Duration] = None,
       applicationName: Option[String] = None,
+      preparedStatementCache: PreparedStatementCache = PreparedStatementCache.default,
       extraProperties: Map[String, String] = Map.empty,
   ) derives JsonSchema {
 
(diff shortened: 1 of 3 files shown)


modules/sql/core/src/main/scala/oxygen/sql/DbConfig.scala
```diff
@@ -69,13 +71,47 @@ object DbConfig {
...
-      ).flatten ++ extraProperties.toSeq
+      ).flatten ++ preparedStatementCache.properties ++ extraProperties.toSeq
 
   }
   object Connection {
     val default: Connection = Connection()
   }
 
+  /**
+    * pgjdbc **server-side prepared-statement caching** knobs. pgjdbc keeps a per-connection cache of server-prepared
```
Kalin-Claude, line 82: **minor** — The header calls the whole block "**server-side** prepared-statement caching knobs," but only `prepareThreshold` concerns server-side prepared plans. `preparedStatementCacheQueries` / `preparedStatementCacheSizeMiB` are pgjdbc's **client-side** statement cache — and this block's own field docs (lines 93–94) plus the `database.md` table both correctly say they "disable **client-side** caching." So the header contradicts its own fields.

Concrete confusion: an operator reading "server-side … caching knobs" may set `cacheQueries: 0` believing it tunes state on the Postgres server, when it actually disables the JDBC client's local statement cache. Suggest rewording the header to something like "prepared-statement caching — client-side statement cache + server-prepare threshold" so the framing matches the knobs. (Same wording carries into the PR description and `database.md` prose, which lead with "server-side" throughout.)

---
:robot: claude

Kalin-Claude (commented):
Adversarial standards + correctness pass on the prepared-statement-cache config. One **minor** (doc/terminology) inline; everything else checked out. No blocking findings.

**Cleared suspicions (checked, not problems):**
- **Default params** — `PreparedStatementCache` defaults every field and the new `Connection.preparedStatementCache` field defaults to `.default`. Not a hard-rules §1 violation: this is a JSON-decode config model (the explicit "edge decode models" carve-out), and it matches the established pattern in this very file (`Connection`, `Pool`, `Logging`, `Execution` all default every field). Opt-in, no-op default is deliberate.
- **Field-insertion / positional shift** — `preparedStatementCache` is inserted *before* `extraProperties`. Grepped every `Connection(…)` construction (main + tests): all use named args, so no positional call site silently rebinds. Safe.
- **Precedence** — `… ++ preparedStatementCache.properties ++ extraProperties.toSeq`, then `buildProperties` does `props.put` in order, so `extraProperties` still wins on key collision. Correct and covered by a test.
- **pgjdbc facts** — property names (`prepareThreshold`, `preparedStatementCacheQueries`, `preparedStatementCacheSizeMiB`), the stated defaults (5 / 256 / 5), and `prepareThreshold: 0` disabling server-prepared plans all match pgjdbc. Correct.
- **Test coverage** — default-empty, translation, disable-via-zero, extraProperties precedence, `buildProperties` threading, JSON decode, and omission are all covered. Solid.

---
:robot: claude

PR #318, opened 2026-08-21
OFF-316: Add support to `Atomically` for `not in transaction -> wrap in transaction`, but `already in transaction -> I dont care about savepoint`

Kalin-Claude:
Adds a weaker sibling of `atomically` to the `oxygen-sql` `Atomically` service: **ensure the effect runs inside *a* transaction, but if we're already inside one, don't bother with a nested savepoint — just run it inline.**

- Not in a transaction -> wrap in a transaction (identical to the current outermost `atomically` behavior).
- Already in a transaction/savepoint -> **no-op wrapper**: run the effect directly (no new savepoint, no extra `BEGIN`).

### Changes
- **`oxygen.storage.Atomically`** (trait) — new `ensureAtomicScoped: URIO[Scope, Unit]` + `ensureAtomic[R, E, A]` members (scaladoc'd), a companion `ensureAtomic` `ZIOAspectAtLeastR[Atomically]` aspect mirroring `atomically`, and the trivial `NoOp` pass-through.
- **`oxygen.sql.Database`** — `private[sql] currentConnectionType` that peeks the current `ConnectionState.connectionType` **without** allocating an atomic child (`getAtomicChild` always does).
- **`oxygen.sql.Atomically`** — implemented `ensureAtomicScoped` in both `LiveDB` and `RollbackDB` (branch on the current connection type: `Transactionless` -> delegate to the existing `atomicallyScoped`; already `Transactional` -> `ZIO.unit`), re-exported `ensureAtomic`, and added object-level `ensureAtomic` aspect vals on `LiveDB`/`RollbackDB`. The existing `atomically` methods are untouched (no regression surface).
- **`sql-it`** — new `AtomicallySpec` proving the semantics via the connection-type peek: outermost `ensureAtomic` opens a transaction; nested-in-transaction stays a transaction (NO savepoint); nested-in-savepoint stays that savepoint; and a contrast that nested plain `atomically` DOES open a savepoint.

### Resolved open questions (from the ticket)
- **Naming/shape:** `ensureAtomic` — both a trait method and an aspect val (mirrors `atomically`).
- **Scope:** applies to **both** `LiveDB` and `RollbackDB`.
- **"Already in a transaction":** treats **both** `Transaction` and `Savepoint` as already-transactional.
- **Refactor vs separate:** existing `atomically` kept fully separate; the new path only *delegates* to the shared `atomicallyScoped`.

### Verification
`oxygen-storage` / `oxygen-sql` / `sql-it` compile clean; scalafmt applied; `AtomicallySpec` — 5/5 pass.

Jira: OFF-316

### DIFF
--- a/modules/sql/core/src/main/scala/oxygen/sql/Atomically.scala
+++ b/modules/sql/core/src/main/scala/oxygen/sql/Atomically.scala
@@ -8,6 +8,7 @@ type Atomically = oxygen.storage.Atomically
 object Atomically {
 
   val atomically: ZIOAspectAtLeastR[Atomically] = oxygen.storage.Atomically.atomically
+  val ensureAtomic: ZIOAspectAtLeastR[Atomically] = oxygen.storage.Atomically.ensureAtomic
 
   /**
     * TLDR : Transaction(commit/rollback) -> Savepoint(commit/rollback) -> Savepoint(commit/rollback) -> ...
@@ -34,6 +35,15 @@ object Atomically {
     override def atomically[R, E, A](effect: ZIO[R, E, A]): ZIO[R, E, A] =
       ZIO.scoped { atomicallyScoped *> effect }
 
+    override def ensureAtomicScoped: URIO[Scope, Unit] =
+      db.currentConnectionType.flatMap {
+        case Database.ConnectionState.ConnectionType.Transactionless  => atomicallyScoped
+        case _: Database.ConnectionState.ConnectionType.Transactional => ZIO.unit
+      }
+
+    override def ensureAtomic[R, E, A](effect: ZIO[R, E, A]): ZIO[R, E, A] =
+      ZIO.scoped { ensureAtomicScoped *> effect }
+
   }
   object LiveDB {
 
@@ -41,6 +51,12 @@ object Atomically {
       new ZIOAspectAtLeastR.Impl[Database] {
         override def apply[R <: Database, E, A](effect: ZIO[R, E, A])(using trace: Trace): ZIO[R, E, A] =
           (effect @@ Atomically.atomically).provideSomeLayer[R](LiveDB.layer)
+      }
+
+    val ensureAtomic: ZIOAspectAtLeastR[Database] =
+      new ZIOAspectAtLeastR.Impl[Database] {
+        override def apply[R <: Database, E, A](effect: ZIO[R, E, A])(using trace: Trace): ZIO[R, E, A] =
+          (effect @@ Atomically.ensureAtomic).provideSomeLayer[R](LiveDB.layer)
       }
 
     val layer: URLayer[Database, Atomically] =
@@ -74,6 +90,15 @@ object Atomically {
     override def atomically[R, E, A](effect: ZIO[R, E, A]): ZIO[R, E, A] =
       ZIO.scoped { atomicallyScoped *> effect }
 
+    override def ensureAtomicScoped: URIO[Scope, Unit] =
+      db.currentConnectionType.flatMap {
+        case Database.ConnectionState.ConnectionType.Transactionless  => atomicallyScoped
+        case _: Database.ConnectionState.ConnectionType.Transactional => ZIO.unit
+      }
+
+    override def ensureAtomic[R, E, A](effect: ZIO[R, E, A]): ZIO[R, E, A] =
+      ZIO.scoped { ensureAtomicScoped *> effect }
+
   }
   object RollbackDB {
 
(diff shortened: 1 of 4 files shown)


modules/sql/core/src/main/scala/oxygen/sql/Atomically.scala
```diff
@@ -34,6 +35,15 @@ object Atomically {
     override def atomically[R, E, A](effect: ZIO[R, E, A]): ZIO[R, E, A] =
       ZIO.scoped { atomicallyScoped *> effect }
 
+    override def ensureAtomicScoped: URIO[Scope, Unit] =
```
Kalin-Claude, line 38: **minor** — this `ensureAtomicScoped` body is byte-identical to `RollbackDB.ensureAtomicScoped` (recurs at ~L93). Unlike `atomicallyScoped`, which genuinely differs between the two impls (commit vs. rollback), these two branch-on-connection-type bodies are the same, so the duplication is new rather than inherited. Could be a shared private helper (e.g. `def ensureAtomicScopedFrom(atomically: URIO[Scope, Unit])`), or — since `ensureAtomic[R,E,A]` is also identical across `LiveDB`/`RollbackDB`/`NoOp` — a default `ensureAtomic` on the trait defined in terms of `ensureAtomicScoped`. Optional: it mirrors how the existing `atomically` is already hand-duplicated, so consistent-with-existing if you'd rather leave it.

---
:robot: claude

modules/sql/it-test/src/test/scala/oxygen/sql/AtomicallySpec.scala
```diff
@@ -0,0 +1,55 @@
...
+    ZIO.serviceWithZIO[Database](_.currentConnectionType)
+
+  private def isSavepoint(ct: Database.ConnectionState.ConnectionType.Any): Boolean =
+    ct match {
+      case _: Database.ConnectionState.ConnectionType.Savepoint => true
+      case _                                                    => false
+    }
+
+  override def testSpec: TestSpec =
+    suite("AtomicallySpec")(
```
Kalin-Claude, line 21: **minor** — all 5 tests drive `Atomically.LiveDB`; `RollbackDB.ensureAtomic` has zero coverage even though the PR description lists it as in-scope ("applies to both `LiveDB` and `RollbackDB`"). Risk is low because the `RollbackDB.ensureAtomicScoped` body is identical to `LiveDB`'s, but the branch-selection claim for `RollbackDB` is currently unproven — one nested-in-transaction case under `RollbackDB.ensureAtomic` asserting `ct == transaction` would close it.

---
:robot: claude

Kalin-Claude (commented):
Reviewer pass (adversarial, standards-focused). The change is correct and idiomatic; I found no blocking issues — no `severe`/`major`. Two `minor` notes inline.

**Semantics verified (cleared suspicions):**
- *Peek TOCTOU* — `ensureAtomicScoped` peeks `currentConnectionType` and then, on the `Transactionless` branch, `atomicallyScoped` independently re-reads via `getAtomicChild`. Because `connectionStateRef` is a `FiberRef` (fiber-local, only mutated by this fiber's own scoped `locallyScoped`), nothing can change the value between the peek and the act within the same fiber, so the two reads stay consistent. Not a bug.
- *Match exhaustiveness* — `ConnectionType.Any` is sealed = `Transactionless | Transactional`, so `Transactionless` + `case _: Transactional` is exhaustive with no stringly/catch-all foul (hard-rules #2). Good.
- *API-compat* — the two new abstract members on the `oxygen.storage.Atomically` trait are implemented in all three in-repo impls (`NoOp`, `LiveDB`, `RollbackDB`); grep finds no other implementor. (They are abstract rather than trait-defaulted, mirroring the existing `atomically`.)
- *No-op scope* — on the already-transactional branch `ensureAtomicScoped` is `ZIO.unit`, so `ZIO.scoped { ensureAtomicScoped *> effect }` allocates no finalizer and leaks nothing; the intentional tradeoff (loss of independent partial-rollback) is documented in the scaladoc.
- *Test conventions* — `AtomicallySpec` matches the `OxygenSpec[Database]` / `layerProvider` / `@@ TestAspect.sequential` house style; braces + single-line `case` RHS are compliant (hard-rules #14).

One stylistic note that is NOT worth a separate thread: the tests assert the fiber connection-state (a legitimate proxy for "did a savepoint get pushed") rather than the emitted SQL. That's a reasonable, lightweight choice — flagging only so it's a conscious one.

---
:robot: claude

PR #320, opened 2026-08-21
OFF-317: Add inline configurability to query timeouts

Kalin-Claude:
Adds a per-effect query-timeout override, modeled on the existing `Database.withQueryLogLevel` / `ModifyLogConfig` aspect.

## Changes
- **`DbConfig.Execution`** gains `queryTimeout: Option[Duration] = None` (`None` = no timeout — current behavior, explicit).
- **`Database`**: `executionConfig` promoted from a plain `val` to a `FiberRef[DbConfig.Execution]`, so it can be overridden fiber-locally (symmetry with `logConfigRef`). `make`, `baseLayer`, and the `executionConfig` accessor updated.
- **`Database.withQueryTimeout(duration)`** (+ `.none`) — a `ZIOAspectAtLeastR[Database]` via a private `ModifyExecutionConfig`, an exact mirror of `withQueryLogLevel`:
  ```scala
  longRunningDbMigrationEffect @@ Database.withQueryTimeout(1.hour)
  ```
- **`PreparedStatement.prepare`** applies the timeout via JDBC `setQueryTimeout`, rounding the `Duration` **up** to whole seconds so a sub-second timeout never collapses to `0` (= no limit); non-positive → left unset. Per-statement, so no reset finalizer needed.
- **it-test `QueryTimeoutSpec`** — default is `None`; `pg_sleep(30) @@ withQueryTimeout(1.second)` aborts; `pg_sleep(1) @@ withQueryTimeout(30.seconds)` succeeds.

## Verification
- `oxygen-sql/compile` + `sql-it/Test/compile` green.
- `QueryTimeoutSpec` — 3/3 pass against a real Postgres container.
- scalafmt clean.

Chose JDBC `setQueryTimeout` over `SET LOCAL statement_timeout` because it needs no surrounding transaction (migrations aren't always inside `Atomically`). A generic `settingsUpdate(...)` wrapper and ms-precision `SET LOCAL` remain possible follow-ups.

Jira: OFF-317

### DIFF
--- a/modules/sql/core/src/main/scala/oxygen/sql/query/PreparedStatement.scala
+++ b/modules/sql/core/src/main/scala/oxygen/sql/query/PreparedStatement.scala
@@ -213,6 +213,19 @@ object PreparedStatement {
           }
         }
         .mapError(QueryError.JDBCError("setting fetch size", _))
+
+      exeCfg <- Database.executionConfig
+      _ <- ZIO
+        .foreachDiscard(queryTimeoutSeconds(exeCfg.queryTimeout)) { seconds => ZIO.attempt { rawPS.setQueryTimeout(seconds) } }
+        .mapError(QueryError.JDBCError("setting query timeout", _))
     } yield ps).uninterruptible.mapError(QueryError(ctx, _))
+
+  /**
+    * JDBC `setQueryTimeout` takes whole seconds, where `0` means "no timeout". Convert a [[Duration]] by
+    * rounding UP to the next whole second so that a sub-second timeout never silently becomes `0` (= no
+    * limit). A non-positive duration yields [[None]] (leaving the JDBC default of no timeout in place).
+    */
+  private def queryTimeoutSeconds(timeout: Option[Duration]): Option[Int] =
+    timeout.collect { case d if d.toNanos > 0L => math.min(Int.MaxValue.toLong, (d.toNanos + 999999999L) / 1000000000L).toInt }
 
 }
--- a/modules/sql/it-test/src/test/scala/oxygen/sql/QueryTimeoutSpec.scala
+++ b/modules/sql/it-test/src/test/scala/oxygen/sql/QueryTimeoutSpec.scala
@@ -1,0 +1,40 @@
+package oxygen.sql
+
+import oxygen.predef.test.*
+import oxygen.sql.query.{QueryContext, QueryO}
+import oxygen.sql.schema.RowRepr
+import zio.*
+
+object QueryTimeoutSpec extends OxygenSpec[Database] {
+
+  // `pg_sleep(n)` blocks for `n` seconds server-side; `::text` makes its `void` result decodable.
+  private def sleep(seconds: Int): ZIO[Database, Throwable, String] =
+    QueryO
+      .simple[String]("PgSleep", QueryContext.QueryType.Select)(RowRepr.string.decoder)(s"SELECT pg_sleep($seconds)::text")
+      .execute()
+      .single
+
+  override def testSpec: TestSpec =
+    suite("QueryTimeoutSpec")(
+      test("queryTimeout defaults to None") {
+        assertTrue(DbConfig.Execution.default.queryTimeout.isEmpty)
+      },
+      test("withQueryTimeout aborts a query that exceeds the timeout") {
+        for {
+          exit <- (sleep(30) @@ Database.withQueryTimeout(1.second)).exit
+        } yield assertTrue(exit.isFailure)
+      },
+      test("a query that completes within the timeout succeeds") {
+        for {
+          res <- sleep(1) @@ Database.withQueryTimeout(30.seconds)
+        } yield assertTrue(res == "")
+      },
+    ) @@ TestAspect.withLiveClock @@ TestAspect.timeout(1.minute) @@ TestAspect.sequential
+
+  override def layerProvider: LayerProvider[R] =
+    LayerProvider.provideShared[Env](
+      Helpers.testContainerLayer,
+      Helpers.databaseLayer,
+    )
+
+}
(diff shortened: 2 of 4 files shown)


Kalin-Claude (commented):
Adversarial standards + correctness pass on the inline query-timeout work. Scope: the 4 files in the diff.

**Overall:** clean, well-modeled change — `ModifyExecutionConfig` / `withQueryTimeout` are an exact mirror of the existing `ModifyLogConfig` / `withQueryLogLevel`, and the `FiberRef` promotion of `executionConfig` is transparent to the existing readers (the `Database.executionConfig` accessor and the streaming reads in `PreparedStatement`). No blocking findings — everything below is `minor`.

**Suspicions I chased and cleared (so they aren't re-litigated):**
- `DbConfig.Execution.queryTimeout: Option[Duration] = None` is a default param on a `case class` field — normally a hard-rule foul, but this is the sanctioned *edge decode model* carve-out: `Execution derives JsonSchema` and both sibling fields (`bufferChunkSize`, `bufferNumChunks`) already default so a missing config key decodes. Consistent, not a violation.
- No reset finalizer for `setQueryTimeout` (unlike `setFetchSizeScoped`) is correct — a fresh `PreparedStatement` is created per query, so the timeout dies with the statement. Body's reasoning checks out.
- `FiberRef.locallyWith` restore semantics are inherited from the proven `logConfigRef` path.

The three inline notes are all optional polish; the test-coverage one is the most worth doing.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/query/PreparedStatement.scala
```diff
@@ -213,6 +213,19 @@ object PreparedStatement {
...
+        .mapError(QueryError.JDBCError("setting query timeout", _))
     } yield ps).uninterruptible.mapError(QueryError(ctx, _))
 
+  /**
+    * JDBC `setQueryTimeout` takes whole seconds, where `0` means "no timeout". Convert a [[Duration]] by
+    * rounding UP to the next whole second so that a sub-second timeout never silently becomes `0` (= no
+    * limit). A non-positive duration yields [[None]] (leaving the JDBC default of no timeout in place).
+    */
+  private def queryTimeoutSeconds(timeout: Option[Duration]): Option[Int] =
+    timeout.collect { case d if d.toNanos > 0L => math.min(Int.MaxValue.toLong, (d.toNanos + 999999999L) / 1000000000L).toInt }
```
Kalin-Claude, line 229: **minor** — the overflow clamp is half-applied. `math.min(Int.MaxValue.toLong, …)` guards the final `Int` cast, but the intermediate `(d.toNanos + 999999999L)` overflows `Long` *before* that guard for very large durations: e.g. a ~292-year timeout has `toNanos` just under `Long.MaxValue`, `+ 999999999L` wraps to negative, `/ 1000000000` stays negative, `math.min` passes it through, and `setQueryTimeout(<negative>)` throws `SQLException` — so *every* query fails. `Duration.Infinity` hits the same trap (its `toNanos` is at/over the ceiling — either the add overflows or `toNanos` itself throws `ArithmeticException`), which is a plausible "no limit" value a caller might pass instead of `withQueryTimeout.none`. Clamp on seconds to stay overflow-free, e.g. work from `d.getSeconds` + a `d.getNano > 0` round-up rather than `toNanos`.

---
:robot: claude

modules/sql/it-test/src/test/scala/oxygen/sql/QueryTimeoutSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  override def testSpec: TestSpec =
+    suite("QueryTimeoutSpec")(
+      test("queryTimeout defaults to None") {
+        assertTrue(DbConfig.Execution.default.queryTimeout.isEmpty)
+      },
+      test("withQueryTimeout aborts a query that exceeds the timeout") {
+        for {
+          exit <- (sleep(30) @@ Database.withQueryTimeout(1.second)).exit
+        } yield assertTrue(exit.isFailure)
```
Kalin-Claude, line 25: **minor** — `assertTrue(exit.isFailure)` only proves the effect failed, not that it failed *because of the timeout*. A regression where the query dies for an unrelated reason (decode error, connection fault, SQL typo) would keep this test green. Assert on the actual cause — that the failure is the JDBC query-timeout `QueryError` — so the test defends the behavior it names.

---
:robot: claude

modules/sql/it-test/src/test/scala/oxygen/sql/QueryTimeoutSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+      .simple[String]("PgSleep", QueryContext.QueryType.Select)(RowRepr.string.decoder)(s"SELECT pg_sleep($seconds)::text")
+      .execute()
+      .single
+
+  override def testSpec: TestSpec =
+    suite("QueryTimeoutSpec")(
+      test("queryTimeout defaults to None") {
+        assertTrue(DbConfig.Execution.default.queryTimeout.isEmpty)
+      },
+      test("withQueryTimeout aborts a query that exceeds the timeout") {
```
Kalin-Claude, line 22: **minor** — coverage gap on the subtlest part of the PR. `queryTimeoutSeconds`' "round sub-second UP so it never collapses to 0 (= no limit)" is the whole point of the ceil logic, and it's the one branch untested — both timeout tests use whole seconds (`1.second`, `30.seconds`). A discriminating case would be `sleep(2) @@ withQueryTimeout(500.millis)` expecting an abort: if the rounding ever regresses to floor/`0`, that becomes "no limit" and the query runs the full 2s and succeeds, flipping the test. Add it to lock the guarantee the scaladoc promises.

---
:robot: claude

PR #321, opened 2026-08-21
OFF-328: Add `on conflict` support to query dsl

Kalin-Claude:
Adds `ON CONFLICT` (upsert) support to the hand-written query DSL, targeting the table's primary key — closing the gap where `ON CONFLICT` was only available in the generated CRUD layer.

## What

- `into(v).onConflictDoNothing` → `ON CONFLICT (pk) DO NOTHING`
- `into(v).onConflictDoUpdate` → `ON CONFLICT (pk) DO UPDATE SET <non-pk> = EXCLUDED.<non-pk>` (falls back to `DO NOTHING` when the table has no non-PK columns)
- Works on both `Q.insert[A]` and `Q.insert.fromSelect[A]`.

## How

- New `OnConflictPart` model, parsed as an optional suffix on the `into(...)` generator in `IntoPart`.
- Threaded through `ParsedQuery.InsertQuery.{Basic,FromSelect}`; emitted by a new `FragmentBuilder.onConflict` (column names resolved from the `TableRepr` at runtime, matching `TableCompanion.upsert`).
- Batch path is unaffected — the `ON CONFLICT …` suffix lands after `VALUES` (as `batchOptimizedUpsert` already relies on).

## Scope

PK-targeted forms only. Explicit non-PK conflict targets, `ON CONSTRAINT`, partial-index `WHERE`, and custom `DO UPDATE SET … WHERE` are deferred as follow-ups.

## Tests

- Compile-time `@compile` snapshots in `queries.scala` (the macro gate).
- Postgres integration test in `CustomQuerySpec` (DO NOTHING keeps the row, DO UPDATE overwrites) + generated-SQL shape assertions — passing locally against a real container.
- Docs: `docs/docs/sql/queries.md`.

Jira: OFF-328

### DIFF
--- a/modules/sql/core/src/main/scala/oxygen/sql/generic/generation/FragmentBuilder.scala
+++ b/modules/sql/core/src/main/scala/oxygen/sql/generic/generation/FragmentBuilder.scala
@@ -541,6 +541,34 @@ final case class FragmentBuilder(inputs: List[InputPart])(using Quotes) {
       valuesFrag,
     )
 
+  /**
+    * `ON CONFLICT (pk...) DO NOTHING | DO UPDATE SET <non-pk> = EXCLUDED.<non-pk>`.
+    * The conflict target is the table's primary-key columns, and `DO UPDATE` sets every non-pk column
+    * to its `EXCLUDED` value -- falling back to `DO NOTHING` when there are no non-pk columns (a bare
+    * `DO UPDATE SET` with no assignments is not valid SQL). The column names are resolved from the
+    * table's `TableRepr` at runtime, mirroring the string built by `TableCompanion.upsert`.
+    */
+  def onConflict(oc: OnConflictPart, tableRepr: TypeclassExpr.TableRepr)(using Quotes): ParseResult[GeneratedFragment] = {
+    val sqlExpr: Expr[String] =
+      oc match {
+        case OnConflictPart.DoNothing =>
+          '{
+            val pkNames: ArraySeq[String] = ${ tableRepr.expr }.pk.rowRepr.columns.columns.map(_.name)
+            "\n    ON CONFLICT (" + pkNames.mkString(", ") + ")\n    DO NOTHING"
+          }
+        case OnConflictPart.DoUpdate =>
+          '{
+            val pkNames: ArraySeq[String] = ${ tableRepr.expr }.pk.rowRepr.columns.columns.map(_.name)
+            val npkNames: ArraySeq[String] = ${ tableRepr.expr }.npk.rowRepr.columns.columns.map(_.name)
+            val target: String = "\n    ON CONFLICT (" + pkNames.mkString(", ") + ")"
+            if npkNames.isEmpty then target + "\n    DO NOTHING"
+            else target + "\n    DO UPDATE\n    SET " + npkNames.map { n => n + " = EXCLUDED." + n }.mkString(",\n        ")
+          }
+      }
+
+    ParseResult.success(GeneratedFragment.sql(sqlExpr))
+  }
+
   private def setPart(s: SetPart.SetExpr)(using ParseContext, GenerationContext, Quotes): ParseResult[GeneratedFragment] = {
     val setTargetRowRepr: TypeclassExpr.RowRepr = s.fieldToSetExpr.rowRepr
     val setTargetColumnNames: Expr[ArraySeq[String]] = setTargetRowRepr.columns.exprSeqNames
--- a/modules/sql/core/src/main/scala/oxygen/sql/generic/model/ParsedQuery.scala
+++ b/modules/sql/core/src/main/scala/oxygen/sql/generic/model/ParsedQuery.scala
@@ -69,6 +69,7 @@ private[sql] object ParsedQuery extends Parser[Term, ParsedQuery] {
           fragmentBuilder = FragmentBuilder(inputs)
           insertFrag <- fragmentBuilder.insert(insert)
           valuesFrag <- fragmentBuilder.values(insert, into)
+          onConflictFrag <- into.onConflict.traverse(fragmentBuilder.onConflict(_, insert.tableRepr)).map(GeneratedFragment.option)
           returningFrag <- fragmentBuilder.ret(ret, "              ")
           frag = GeneratedFragment.of(
             insertFrag,
@@ -76,6 +77,7 @@ private[sql] object ParsedQuery extends Parser[Term, ParsedQuery] {
             '{ ${ insert.tableRepr.expr }.rowRepr.columns.`a, b, c` },
             ")",
             valuesFrag,
+            onConflictFrag,
             GeneratedFragment.option(
               returningFrag.map { ret =>
                 GeneratedFragment.of(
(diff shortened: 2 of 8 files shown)


modules/sql/core/src/main/scala/oxygen/sql/generic/model/ParsedQuery.scala
```diff
@@ -69,13 +69,15 @@ private[sql] object ParsedQuery extends Parser[Term, ParsedQuery] {
           fragmentBuilder = FragmentBuilder(inputs)
           insertFrag <- fragmentBuilder.insert(insert)
           valuesFrag <- fragmentBuilder.values(insert, into)
+          onConflictFrag <- into.onConflict.traverse(fragmentBuilder.onConflict(_, insert.tableRepr)).map(GeneratedFragment.option)
```
Kalin-Claude, line 72: **minor** — the `ON CONFLICT` clause is threaded into the generated SQL fragment here, but not into the human-readable `show`. `Basic.show` (this class, ~L52-55) and `FromSelect.show` (~L110-113) still render only `INSERT … VALUES …` / `INSERT … <select>`, and that `show` is what `@compile(debug = true)` prints as `--- parsed query ---` (`CompileMacros.scala:28`). Concrete effect: `@compile(debug = true) upsertPerson` prints a query indistinguishable from a plain insert — the exact clause you'd be debugging is invisible. Recurs in both insert `show`s. The runtime SQL is unaffected (built from fragments), so this is display-only.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/generic/generation/FragmentBuilder.scala
```diff
@@ -541,6 +541,34 @@ final case class FragmentBuilder(inputs: List[InputPart])(using Quotes) {
...
+    * to its `EXCLUDED` value -- falling back to `DO NOTHING` when there are no non-pk columns (a bare
+    * `DO UPDATE SET` with no assignments is not valid SQL). The column names are resolved from the
+    * table's `TableRepr` at runtime, mirroring the string built by `TableCompanion.upsert`.
+    */
+  def onConflict(oc: OnConflictPart, tableRepr: TypeclassExpr.TableRepr)(using Quotes): ParseResult[GeneratedFragment] = {
+    val sqlExpr: Expr[String] =
+      oc match {
+        case OnConflictPart.DoNothing =>
+          '{
+            val pkNames: ArraySeq[String] = ${ tableRepr.expr }.pk.rowRepr.columns.columns.map(_.name)
```
Kalin-Claude, line 556: **minor** — guard-parity with the CRUD path: `TableCompanion.upsert` and `insertOrDoNothing` both open with `if pkCols.isEmpty then throw …` before building `ON CONFLICT (…)`. `onConflict` here has no such guard, so a `TableRepr` with zero PK columns would emit `ON CONFLICT ()` — a Postgres syntax error at execution rather than a clear failure at build. This is lower-confidence (I couldn't confirm a PK-less table is even reachable through `Q.insert[A]`), but the sibling CRUD path evidently considered it worth guarding, and parity is a one-liner.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/generic/model/part/IntoPart.scala
```diff
@@ -26,12 +28,18 @@ object IntoPart extends MapChainParser[IntoPart] {
       mapAAFC <- AppliedAnonFunctCall.parseTyped[T.InsertValues](term, "map function").ignore
       _ <- mapAAFC.funct.parseEmptyParams
       mapFunctName <- functionNames.mapOrFlatMap.parse(mapAAFC.nameRef).unknownAsError
-      (_, argTerm) <- mapAAFC.lhs match { // TODO (KR) : validate `intoIdent`
+      // peel an optional `.onConflictDoNothing` / `.onConflictDoUpdate` suffix off the `into(...)` call
+      (intoTerm, onConflict) = mapAAFC.lhs match {
+        case Select(inner, "onConflictDoNothing") => (inner, Option(OnConflictPart.DoNothing))
```
Kalin-Claude, line 33: **nit** — `Option(OnConflictPart.DoNothing)` / `Option(OnConflictPart.DoUpdate)`: the argument is a statically non-null enum case, so `Option.apply`'s null check is dead and the call reads as "this might be null." Idiomatic here is `OnConflictPart.DoNothing.some` (Oxygen predef, used throughout the codebase) or `Some(...)`.

---
:robot: claude

modules/sql/it-test/src/test/scala/oxygen/sql/queries.scala
```diff
@@ -217,6 +217,38 @@ object queries {
...
+    for {
+      (_, into) <- Q.insert.fromSelect[Note]
+      p <- Q.select[Person]
+      _ <- into(
+        Note(
+          UUID.randomUUID(),
+          p.id,
+          mkSqlString(Q.const("Adding note for person "), p.first, " ", p.last, " : everyone"),
+        ),
+      ).onConflictDoNothing
```
Kalin-Claude, line 249: **minor** (coverage) — this `fromSelect` + `onConflict` query is only shape-asserted in `CustomQuerySpec` via `.ctx.sql.contains("ON CONFLICT (id)")`; unlike the `Basic` cases it's never executed against the container. More notably, the PR body + docs claim it "composes with `batchOptimizedInsert` (the `ON CONFLICT …` suffix rides along after `VALUES`)" — and that claim has **no** test. It rests entirely on `BatchOptimizedInsert.SqlParts.unsafeParse`'s regex isolating the `VALUES` tuple; that holds today (I verified), but a future tweak to either the emitted whitespace or the regex would silently mis-split with no failing test. A `BatchOptimizedInsert.unsafeParse(<onConflict DSL query>)` round-trip asserting the expanded multi-row SQL would lock the documented behavior down.

---
:robot: claude

Kalin-Claude (commented):
Adversarial standards + correctness pass on the `ON CONFLICT` DSL work. The core is sound: the runtime-emitted strings in `FragmentBuilder.onConflict` are byte-for-byte parity with `TableCompanion.upsert`/`insertOrDoNothing`, the empty-npk `DO NOTHING` fallback is correct, and fragment ordering (`VALUES` → `ON CONFLICT` → `RETURNING`) is valid for both `Basic` and `fromSelect`.

**Suspicions I chased and cleared:**
- *Suffix peeling* (`IntoPart`): the `Select(inner, "onConflict…")` peel sits correctly on the map/flatMap receiver; the no-suffix `into(i)` falls through to the `other` branch as an `Apply(Select(_,"apply"),…)`. Both `Basic` and `fromSelect` route through the same `parse` (FromSelect defers via `IntoPart.map`), so the clause is carried through in both. Good.
- *Batch path*: I traced `BatchOptimizedInsert.SqlParts.unsafeParse`'s regex `(?s)\A(.*)(\([?, \n]+\))(.*)?\z` against a DoUpdate output — the only `[?, \n]`-only paren is the `VALUES (?, …)` tuple (`(id)` and the `SET … EXCLUDED.…` clause don't qualify), so the `ON CONFLICT …` suffix correctly lands in `afterValues` and replicates once. The claim holds — but it is untested (see inline).
- *`UUID.randomUUID()`* in the new `fromSelect` test: established DSL pattern (3 pre-existing on `main`), macro-parsed, not a runtime effect — not a §18 violation.

Findings below are all **minor/nit** — nothing blocking. Leaving as comments rather than approving so the author can weigh the debug-`show` gap and the guard-parity item.

---
:robot: claude

PR #322, opened 2026-08-21
OFF-427: Weird mobile top bar issue (when url is shown/hidden)

Kalin-Claude:
## Problem

On mobile the browser URL bar shows/hides as you scroll, changing the **visible** viewport height. The CSS `vh` unit always resolves to the **large** viewport (URL bar hidden), so any `height: 100vh` element is taller than what's actually visible while the URL bar is showing.

The `HolyGrail` shell used an inline `height: 100vh`. An inline style beats a non-`!important` stylesheet rule, so it overrode `HolyGrail.responsiveSheet`'s below-`md` `height: auto` / `min-height: 100dvh` — the shell got sized to the large viewport and the top grid row (the top bar) was pushed partly off-screen and appeared to jump as the URL bar toggled. (Jira OFF-427.)

## Fix

- **`.dvh` helper** added next to `.vh` in `create/extensions.scala`.
- **`HolyGrail`**: removed both inline `height: 100vh` (grid + center-only shells) and moved the height into a new `HolyGrail.baseSheet` that emits a `height: 100vh; height: 100dvh;` **fallback pair** (modern browsers take the valid `dvh`; older ones fall back to `vh`). Registered in `coreOxygenStyleSheets` **before** `responsiveSheet` so the below-`md` mobile rule still wins. This removes the inline-vs-stylesheet fight entirely.
- **Bonus defect fixed**: `responsiveSheet`'s `min-height` fallback pair was written in the wrong order (`100dvh` then `100vh`), which made modern browsers resolve `min-height` to `100vh`. Flipped to `100vh` then `100dvh`.
- **Audit of other full-viewport `100vh` consumers**:
  - `OxygenStyleSheet` `ModalOverlay` (a stylesheet selector) → emits the `100vh → 100dvh` pair.
  - `Drawer` + `CenteredCard` (inline) → `100.dvh` (bare — inline styles are a single value per property, so no pair is possible; this matches the existing house-style inline `dvh` in `PageMessagesBottomCorner`/`Modal`).
  - The ticket also named `OxygenStyleSheet:391` (a Dropdown "Scrim"); that line does not exist on `main` (only on the `pr-301` branch the ticket was drafted against), so it's out of scope here.

Why the shell uses a stylesheet fallback pair while `Drawer`/`CenteredCard` use bare inline `dvh`: inline CSS in this DSL is a `Map[String, String]` (one value per property), so a fallback pair is only possible in `selector(...)` stylesheet blocks. The shell — the highest-value element and the one the ticket calls out for a fallback — is worth moving to a sheet; the secondary inline consumers follow existing house style.

## Verification

- `oxygen-ui-web/compile` — green.
- `oxygen-ui-web/scalafmt` — clean.
- No tests reference these literals.

## ⚠️ Residual uncertainty

The fix is derived from the CSS + UI-DSL mechanics (confirmed against the source), **but I could not verify it on a real mobile device** in this environment. Worth a quick manual check on a phone that:
1. the top bar no longer clips/jumps as the URL bar shows/hides, and
2. center-scroll containment still behaves (the ticket left this as an open question).

I kept `dvh` (not `svh`) per the ticket's default; `dvh` re-layouts on every URL-bar transition, a known accepted trade-off.

Jira: OFF-427

### DIFF
--- a/modules/ui/web/src/main/scala/oxygen/ui/web/layout/HolyGrail.scala
+++ b/modules/ui/web/src/main/scala/oxygen/ui/web/layout/HolyGrail.scala
@@ -43,7 +43,8 @@ final case class HolyGrail[-Env, +Action, -StateGet, +StateSet <: StateGet](
           Widget.`class`("oxy-holy-grail"),
           //
           display.grid,
-          height := 100.vh,
+          // height comes from `baseSheet` (100vh → 100dvh fallback pair); kept out of the inline
+          // style so the below-md `responsiveSheet` rule can win on mobile (OFF-427).
           width := 100.vw,
           minHeight := 0,
           minWidth := 0,
@@ -66,7 +67,7 @@ final case class HolyGrail[-Env, +Action, -StateGet, +StateSet <: StateGet](
       else
         div(
           Widget.`class`("oxy-holy-grail", "oxy-holy-grail--center-only"),
-          height := 100.vh,
+          // height from `baseSheet` (see grid shell above) — OFF-427.
           width := 100.vw,
           minHeight := 0,
           minWidth := 0,
@@ -301,6 +302,27 @@ object HolyGrail extends WidgetTypes[HolyGrail] {
   def apply(): HolyGrail.Const = empty
 
   /**
+    * OFF-427: base height for the shell, kept out of the inline style so the below-`md`
+    * [[responsiveSheet]] rule can win on mobile (an inline style beats any non-`!important`
+    * stylesheet rule).
+    *
+    * The `100vh` → `100dvh` pair is a CSS fallback: modern browsers take the later, valid `100dvh`
+    * (tracks the *visible* viewport as the mobile URL bar shows/hides — the bug); older browsers that
+    * don't understand `dvh` drop it and fall back to `100vh`.
+    *
+    * MUST be registered *before* [[responsiveSheet]] in [[oxygen.ui.web.defaults.coreOxygenStyleSheets]]
+    * — equal specificity, media queries add none, so source order decides who wins below `md`.
+    */
+  val baseSheet: StyleSheet =
+    StyleSheet.makeConst("holy-grail-base")(
+      """.oxy-holy-grail {
+        |  height: 100vh;
+        |  height: 100dvh;
+        |}
+        |""".stripMargin,
+    )
+
+  /**
     * W5-T04: below `md`, hide left/right sidebars and force a single-column shell
     * so fixed desktop grids don't blow up mobile. Drawer/hamburger = Round 2 / TopBar todo.
     *
@@ -318,8 +340,8 @@ object HolyGrail extends WidgetTypes[HolyGrail] {
           |    "center"
           |    "bottom-bar" !important;
           |  height: auto;
-          |  min-height: 100dvh;
           |  min-height: 100vh;
+          |  min-height: 100dvh;
           |  width: 100%;
           |}
           |.oxy-holy-grail-left,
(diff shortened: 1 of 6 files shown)


modules/ui/web/src/main/scala/oxygen/ui/web/layout/HolyGrail.scala
```diff
@@ -43,7 +43,8 @@ final case class HolyGrail[-Env, +Action, -StateGet, +StateSet <: StateGet](
           Widget.`class`("oxy-holy-grail"),
           //
           display.grid,
-          height := 100.vh,
+          // height comes from `baseSheet` (100vh → 100dvh fallback pair); kept out of the inline
+          // style so the below-md `responsiveSheet` rule can win on mobile (OFF-427).
           width := 100.vw,
           minHeight := 0,
```
Kalin-Claude, line 49: **major** — the below-`md` fix only half-lands. Removing the inline `height` correctly lets `responsiveSheet`'s `height: auto` win, but this inline `min-height: 0` (specificity 1,0,0,0) still shadows `responsiveSheet`'s **non-`!important`** `.oxy-holy-grail { min-height: 100dvh }` (0,0,1,0) — the exact inline-beats-stylesheet mechanism the PR description invokes for `height`. So below `md` the shell resolves to `height: auto; min-height: 0` and collapses to **content** height instead of filling the visible viewport: on any HolyGrail mobile page whose content is shorter than the viewport, the shell stops short (background gap / bottom-bar rides up).

Same inline `min-height: 0` is on the center-only shell (line 72), so it recurs there. Corollary: the "bonus" `min-height: 100vh; 100dvh` reorder in `responsiveSheet` has no observable effect on the shell — that `min-height` never wins against this inline `0`.

To actually land the intended below-`md` `min-height: 100dvh`: override it in `responsiveSheet` with `min-height: 100dvh !important` (matching the `!important` already on the grid-template rules there), or move `min-height` off the inline style below `md`. Confirmed from source: `DOMElement.cssStr` renders inline styles from a `Map`, and inline beats a non-`!important` sheet rule. Warrants the same on-device check the PR flags.

---
:robot: claude

Kalin-Claude (commented):
Reviewer pass (adversarial, standards-focused). **Not approving** — review-only mandate for this turn, and there's one blocking finding below.

The headline `100vh → 100dvh` mechanics are sound, but the below-`md` fix only **half** lands: the `min-height: 100dvh` it leans on is shadowed by an inline `min-height: 0` that the PR left in place (inline comment on `HolyGrail.scala`). That's the same inline-beats-stylesheet trap the description says it "removed entirely" — removed for `height`, not for `min-height`.

**Verified against source (cleared suspicions), all correct:**
- Stylesheet-selector blocks keep duplicate property keys — `StyleSheetElement.elems: Seq`, and `StyleSheet.compile` groups leaves by selector with no key-dedup, sorted by `loc.line` — so ModalOverlay's `height: 100vh; height: 100dvh` pair renders in that order (vh line sorts before dvh). ✅
- Inline widget styles collapse via `.toMap` (`DOMElement.cssAttrMap`) — one value per property — so the bare inline `dvh` on `Drawer`/`CenteredCard` is genuinely the only option; that justification holds. ✅
- `baseSheet` registered before `responsiveSheet`; equal specificity (`.oxy-holy-grail`, media adds none) → source order → `responsiveSheet` wins below `md` for `height`. ✅
- Bonus reorder `min-height: 100vh; 100dvh;` is correct *in isolation* — but see the inline-`min-height:0` finding: on the shell that rule never wins, so the reorder has no observable effect.

**Nit (not inline) — branch is behind `main`:** `coreOxygenStyleSheets` on `main` now has `TopBar.responsiveSheet` between `HolyGrail.responsiveSheet` and `Motion.sheet` (added in `c529f24b`, OXY-151); it's absent from this PR's base, so the branch predates it. Rebase before merge and keep `baseSheet` ordered *before* `HolyGrail.responsiveSheet`. (`TopBar.responsiveSheet` targets `.oxy-topbar*`, not `.oxy-holy-grail`, so no ordering interaction — merge should be clean, just stale.)

---
:robot: claude

PR #323, opened 2026-08-21
OFF-341: Add compatibility spec for DB Schema

Kalin-Claude:
Adds a pure, DB-free **compatibility spec for the DB migration schema** — the DB analogue of the HTTP-schema `oxygen.schema.compat` lattice, and the richer, type-aware counterpart to the coarse `MigrationGenerator.classify` gate.

## What

New package `oxygen.sql.migration.compat` (module `oxygen-sql-migration`):

- **`DbCompatibilitySpec.compare(from: MigrationState, to: MigrationState): DbComparisonResult`** — the pure entry point. Diffs extensions, schemas, and every table (added / removed / changed), and within a changed table its columns, primary key, foreign keys, and indices.
- **`TypeComparison`** — per-`Column.Type` verdict (`Same`/`Widening`/`Narrowing`/`Changed`/`ArrayElem`), fail-closed: only `SMALLINT→INT→BIGINT` and `REAL→DOUBLE PRECISION` widen (compatible); their reverses narrow; `Array` delegates to its element; every other transition is incompatible.
- **`DbComparisonResult`** (+ nested `TableComparison`/`ColumnComparison`/`ConstraintComparison`) — reuses `FromToValues`/`AddedRemovedBoth` from `oxygen.schema.compat`; exposes `toIndentedString`/`pruned`/`isDifferent` and **collapses to the binary `MigrationCompatibility`** the harness already consumes, so the rich diagnostics never disagree with the coarse gate.

## Notes

- **Spec-only** — this is a diagnostic lens over two schema states, not a new gate; `MigrationCheck`/`MigrationGenerator` are untouched (a clean follow-up could route them through the richer result).
- Rules **agree with `MigrationGenerator.classifyOne`** where they overlap; the one addition is column-type widening/narrowing (which the flag ignores). The empty-table constraint exception falls out **structurally** (constraints on a freshly-added table live in the `added` bucket).
- Pure **unit tests, no DB** (20 tests) covering each rule; docs added to `docs/docs/sql/migrations.md`.

## Verification

- `oxygen-sql-migration/Test/compile` — green
- `oxygen-sql-migration/testOnly *DbCompatibilitySpecSpec` — 20 passed, 0 failed
- `scalafmt` (Compile + Test) — clean

Jira: OFF-341

### DIFF
--- a/modules/sql/migration/src/main/scala/oxygen/sql/migration/compat/DbComparisonResult.scala
+++ b/modules/sql/migration/src/main/scala/oxygen/sql/migration/compat/DbComparisonResult.scala
@@ -1,0 +1,163 @@
+package oxygen.sql.migration.compat
+
+import oxygen.predef.core.*
+import oxygen.schema.compat.{AddedRemovedBoth, FromToValues}
+import oxygen.sql.migration.model.*
+import oxygen.sql.migration.model.EntityRef.*
+import oxygen.sql.migration.persistence.model.MigrationCompatibility
+import oxygen.sql.schema.Column
+
+/**
+  * The structural diff of two [[MigrationState]]s (`from` -> `to`), the DB analogue of
+  * `oxygen.schema.compat.ComparisonResult`. It reuses that package's generic `FromToValues` /
+  * `AddedRemovedBoth` lattice, layers on DB-specific compatibility rules, and collapses to the
+  * binary [[MigrationCompatibility]] the migration harness already understands -- while retaining
+  * per-column diagnostics via [[toIndentedString]].
+  *
+  * The rules agree with `MigrationGenerator.classifyOne` everywhere they overlap; the one addition
+  * is per-column type widening/narrowing (which the coarse gate does not model). The empty-table
+  * exception falls out structurally: constraints on a freshly-`added` table sit in the added bucket
+  * (compatible), and only constraints on a table present in `both` states face the strict rules.
+  */
+final case class DbComparisonResult(
+    extensions: AddedRemovedBoth.Many[String, String],
+    schemas: AddedRemovedBoth.Many[SchemaRef, SchemaRef],
+    tables: AddedRemovedBoth.Many[TableState, DbComparisonResult.TableComparison],
+) {
+
+  import DbComparisonResult.worst
+
+  lazy val compatibility: MigrationCompatibility =
+    worst(
+      Seq(
+        extensions.added.map(_ => MigrationCompatibility.BackwardsCompatible), // CREATE EXTENSION
+        extensions.removed.map(_ => MigrationCompatibility.BackwardsCompatible), // dropping an extension is not itself data-breaking
+        schemas.added.map(_ => MigrationCompatibility.BackwardsCompatible), // CREATE SCHEMA
+        schemas.removed.map(_ => MigrationCompatibility.Incompatible), // DROP SCHEMA
+        tables.added.map(_ => MigrationCompatibility.BackwardsCompatible), // CREATE TABLE
+        tables.removed.map(_ => MigrationCompatibility.Incompatible), // DROP TABLE
+        tables.both.map(_.compatibility),
+      ).flatten,
+    )
+
+  def isCompatible: Boolean = compatibility == MigrationCompatibility.BackwardsCompatible
+
+  def isDifferent: Boolean =
+    extensions.added.nonEmpty || extensions.removed.nonEmpty ||
+      schemas.added.nonEmpty || schemas.removed.nonEmpty ||
+      tables.added.nonEmpty || tables.removed.nonEmpty || tables.both.exists(_.isDifferent)
+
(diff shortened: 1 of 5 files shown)


modules/sql/migration/src/main/scala/oxygen/sql/migration/compat/DbComparisonResult.scala
```diff
@@ -0,0 +1,163 @@
...
+      indices: AddedRemovedBoth.Many[IndexState, ConstraintComparison[IndexState]],
+  ) {
+
+    lazy val compatibility: MigrationCompatibility =
+      worst(
+        Seq(
+          columns.added.map(c => if c.nullable then MigrationCompatibility.BackwardsCompatible else MigrationCompatibility.Incompatible), // add nullable ok / non-nullable breaks
+          columns.removed.map(_ => MigrationCompatibility.Incompatible), // DROP COLUMN
+          columns.both.map(_.compatibility),
+          Seq(if primaryKey.isDifferent then MigrationCompatibility.Incompatible else MigrationCompatibility.BackwardsCompatible), // PK change on existing table
```
Kalin-Claude, line 89: **minor** — this PK-change rule is a genuine divergence from the coarse gate, not covered by the scaladoc claim on L17 ("the one addition is per-column type widening/narrowing") nor the PR body / `migrations.md` line that "the rich diagnostics never disagree with the coarse gate."

`MigrationGenerator.classify` has **no** primary-key case at all: a PK alteration is rejected upstream by `StateDiffer` as a hard `StateDiffError` (`InvalidPrimaryKeyAlteration`, `StateDiffer.scala:68`) and never reaches `classify`. Concretely, for two states differing **only** in their PK column set (e.g. PK `{id}` → `{k}`, both columns already present and non-null), `compare(from, to).compatibility == Incompatible` while the generator pipeline produces no migration file at all (it errors). Not dangerous — both are non-permissive — but the "never disagree" / "one addition" wording overstates the agreement: PK-change is a **second** behavior the coarse gate doesn't model. Either drop the PK rule to truly mirror the gate, or document it as an explicit second addition (like widening).

---
:robot: claude

modules/sql/migration/src/main/scala/oxygen/sql/migration/compat/DbComparisonResult.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  import DbComparisonResult.worst
+
+  lazy val compatibility: MigrationCompatibility =
+    worst(
+      Seq(
+        extensions.added.map(_ => MigrationCompatibility.BackwardsCompatible), // CREATE EXTENSION
+        extensions.removed.map(_ => MigrationCompatibility.BackwardsCompatible), // dropping an extension is not itself data-breaking
+        schemas.added.map(_ => MigrationCompatibility.BackwardsCompatible), // CREATE SCHEMA
+        schemas.removed.map(_ => MigrationCompatibility.Incompatible), // DROP SCHEMA
```
Kalin-Claude, line 36: **minor** (test coverage) — `schemas.removed → Incompatible` (this line) and both `extensions` buckets feed `compatibility` but are never exercised: the suite covers only columns / tables / constraints / type. Given the load-bearing "never disagrees with the coarse gate" guarantee, worth adding (a) a `DROP SCHEMA` (and extension add/remove) case, and (b) an oracle test asserting `DbCompatibilitySpec.compare(a, b).compatibility` equals `MigrationGenerator.classify(...)` over a shared set of transitions. That cross-check is the objective way to substantiate the agreement claim — and is exactly what would have surfaced the PK point above / would catch future drift.

---
:robot: claude

Kalin-Claude (commented):
Reviewer pass (fresh review) on the DB-schema compatibility spec. Read the new package against `MigrationGenerator.classifyOne`, `StateDiffer`, the reused `oxygen.schema.compat` lattice, and the model types.

**Standards:** clean. No `var`/`null`/`throw`/`asInstanceOf`/`implicit`/default-params; braces on all templates, braceless single-line `match`, explicit return types, `final case class`es, parameterized `enum` with per-case data. The pervasive `s"…"` in the `toIndentedString` builders is the module's established idiom (72 `s"…"` vs 0 `str"…"` in `sql-migration` main, e.g. `Column.toSql`, `TableState.toIndentedString`) — not flagged.

**Compat lattice — verdicts check out.** Widenings `SMALLINT→INT→BIGINT` / `REAL→DOUBLE PRECISION` are genuinely data-preserving → BC; their reverses → Incompatible; cross-family (incl. `NUMERIC`) fail closed; `Array` delegates to element. The empty-table exception falls out structurally (constraints on an `added` table sit in the added bucket), matching `classifyOne`'s `createdTables` guard. nullable relax/tighten and the `worst` fold (total over the 2-case `MigrationCompatibility`, mirrors `classify`) are correct. FK/index auto-name renames only shift a name alongside a table/column rename, which is already Incompatible structurally — no observable top-level disagreement there.

Two **minor**, non-blocking notes inline. Not approving — this is a review-only turn.

---
:robot: claude

PR #324, opened 2026-08-21
OFF-338: Support explicitly naming FK + IDX

Kalin-Claude:
Wire explicit constraint/index names through `oxygen-sql` table derivation. The downstream plumbing (`ForeignKeyRepr`/`IndexRepr` `explicitName`, `ForeignKeyState`/`IndexState` fallback to auto-names, migration DDL emitting the resolved name) already existed; the four `DeriveTableRepr` sites hard-coding `None` were the only gap.

## Approach

Overloading the annotation constructors (auto vs. named) breaks Scala 3 quote pattern-matching — a class with two constructors makes every `'{ … }` pattern fail to type. Instead, keep the existing annotations **unchanged** (backwards compatible, auto-named) and add single-constructor `.named` subclasses, mirroring the existing `index.unique` nesting:

| Auto-named | Explicitly named |
|---|---|
| `@references[T]` | `@references.named[T]("fk_name")` |
| `@foreignKey[A,B](…)` | `@foreignKey.named[A,B]("fk_name", …)` |
| `@indexed` | `@indexed.named("idx_name")` |
| `@indexed.unique` | `@indexed.unique.named("idx_name")` |
| `@index[A](…)` | `@index.named[A]("idx_name", …)` |
| `@index.unique[A](…)` | `@index.unique.named[A]("idx_name", …)` |

The name is the leading argument, validated at compile time (must be a string literal); a blank name trims to `None` (auto).

## Tests

- **`ExplicitNamingSpec`** (oxygen-sql) — derivation asserts `explicitName` is populated for every named variant (FK/IDX × class/field × unique) and empty for auto.
- **`ExplicitNamingMigrationSpec`** (sql-it) — round-trip asserts the generated DDL contains `ADD CONSTRAINT fk_…` / `CREATE UNIQUE INDEX idx_…` verbatim for named, and auto falls back to generated names.

Docs added to `docs/docs/sql/models.md` (Foreign keys & indices).

## Deferred (follow-up candidates)

Duplicate-name detection and the Postgres 63-byte / identifier-charset checks are left to Postgres at migrate-time (auto-names have no such guard today either).

Jira: OFF-338

### DIFF
--- a/modules/sql/core/src/main/scala/oxygen/sql/generic/typeclass/DeriveTableRepr.scala
+++ b/modules/sql/core/src/main/scala/oxygen/sql/generic/typeclass/DeriveTableRepr.scala
@@ -12,6 +12,12 @@ final class DeriveTableRepr[A: Type](
     instances: Expressions[RowRepr, A],
 )(using quotes: Quotes, fTpe: Type[RowRepr], generic: ProductGeneric[A]) {
 
+  /** Convert an explicit-name annotation argument into `Option[String]`: a blank/whitespace name means "auto-generate" (`None`). */
+  private def explicitNameExpr(nameExpr: Expr[String])(using Quotes): Expr[Option[String]] = {
+    val trimmed: String = nameExpr.valueOrAbort.trim
+    if trimmed.isEmpty then '{ Option.empty[String] } else '{ Some(${ Expr(trimmed) }) }
+  }
+
   private def schemaNameExpr: Expr[String] =
     Expr { generic.annotations.optionalOfValue[schemaName].fold("public")(_.name) }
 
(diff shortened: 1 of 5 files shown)


modules/sql/core/src/main/scala/oxygen/sql/generic/typeclass/DeriveTableRepr.scala
```diff
@@ -12,6 +12,12 @@ final class DeriveTableRepr[A: Type](
     instances: Expressions[RowRepr, A],
 )(using quotes: Quotes, fTpe: Type[RowRepr], generic: ProductGeneric[A]) {
 
+  /** Convert an explicit-name annotation argument into `Option[String]`: a blank/whitespace name means "auto-generate" (`None`). */
+  private def explicitNameExpr(nameExpr: Expr[String])(using Quotes): Expr[Option[String]] = {
+    val trimmed: String = nameExpr.valueOrAbort.trim
+    if trimmed.isEmpty then '{ Option.empty[String] } else '{ Some(${ Expr(trimmed) }) }
```
Kalin-Claude, line 18: **minor** — The blank/whitespace-name -> `None` (auto) branch is a documented public contract (the PR body and `models.md` both promise `.named("")` is equivalent to the plain annotation), but nothing exercises it. `ExplicitNamingSpec` only covers non-blank named vs. bare-annotation auto; neither spec asserts that `@references.named[Parent]("")` (or `"   "`) derives `explicitName.isEmpty`. Adding one case here keeps this branch from silently regressing.

---
:robot: claude

modules/sql/core/src/main/scala/oxygen/sql/generic/typeclass/DeriveTableRepr.scala
```diff
@@ -12,6 +12,12 @@ final class DeriveTableRepr[A: Type](
     instances: Expressions[RowRepr, A],
 )(using quotes: Quotes, fTpe: Type[RowRepr], generic: ProductGeneric[A]) {
 
+  /** Convert an explicit-name annotation argument into `Option[String]`: a blank/whitespace name means "auto-generate" (`None`). */
+  private def explicitNameExpr(nameExpr: Expr[String])(using Quotes): Expr[Option[String]] = {
+    val trimmed: String = nameExpr.valueOrAbort.trim
```
Kalin-Claude, line 17: **nit** — `valueOrAbort` does enforce the "must be a string literal" contract the docs claim, but on a non-literal name (`val n = "x"; @index.named[T](n, …)`) it aborts with the generic quotes message ("Expected a known value…") that doesn't mention the annotation or the literal requirement. A `report.errorAndAbort("<annotation> name must be a string literal", nameExpr)` fallback would make the compile error self-explanatory. Optional given the PR explicitly defers richer name validation.

---
:robot: claude

modules/sql/it-test/src/test/scala/oxygen/sql/migration/ExplicitNamingMigrationSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+  final case class MigChildAuto(@primaryKey id: UUID, parentId: UUID, name: String)
+  object MigChildAuto extends TableCompanion[MigChildAuto, UUID](TableRepr.derived[MigChildAuto])
+
+  /** Genesis migration SQL (empty -> tables), joined for substring assertions. */
+  private def genesisSql(reprs: TableRepr[?]*): String = {
+    val state: MigrationState =
+      MigrationState.fromTables(ArraySeq.from(reprs)) match
+        case Right(s)    => s
+        case Left(error) => throw new RuntimeException(s"could not derive state: $error")
+    val diffs =
```
Kalin-Claude, line 38: **nit** — `throw new RuntimeException(…)` on the `Left` branches deviates from the no-`throw` house idiom. It's test-only diagnostic code for a state that shouldn't happen, so it's defensible, but surfacing the failure through the test framework (e.g. fold the `Either` into an `assertTrue`/failed `TestResult`, or `.getOrElse(fail(...))`) reads more in line with the rest of the suite. Applies to both `Left` arms.

---
:robot: claude

Kalin-Claude (commented):
Reviewed the explicit FK/IDX naming change against the Scala + Oxygen standards. **Verdict: comment only** (nothing blocking).

The core design is sound and I verified it end-to-end:

- **Annotation-subclass approach is correct.** `references`/`foreignKey`/`indexed`/`index` stay single-constructor and backwards-compatible; the `.named` subclasses each add one leading `name: String` and forward `refs*`/`cols*` to the parent ctor. Dropping `final case class` from `references` is safe — it's only ever consumed via quote pattern matching / `allOf[references[?]]` (Expr-level, `asExprOf`), never via `FromExpr`, so case-ness was never load-bearing.
- **Subclass annotations are picked up.** `Annotations.allOf` (modules/general/quoted/.../Annotations.scala:12) widens each annotation term with `asExprOf[Annot]`, so a `@references.named` instance is correctly returned by `allOf[references[?]]` and then matched against the `.named` quote pattern first.
- **Quote-pattern ordering is correct at all four sites** — most-specific (`.unique.named`, `.named`) matched before the plain forms in every `match` (uniqueness, name, and varargs extraction), so no `.named` tree is misclassified as auto.
- **Constructor positions line up.** `$explicitName` is emitted as the first arg of `ForeignKeyRepr(explicitName, …)` / `IndexRepr(explicitName, unique, …)` (schema/ForeignKeyRepr.scala:5, schema/IndexRepr.scala:5), and `None` fallback flows to `ForeignKeyState.autoFKName` (`fk____…`) / `IndexState.autoIdxName` (`idx_u____…`), which is exactly what the migration spec asserts.

Findings are all non-blocking (one `minor`, two `nit`) and inline.

---
:robot: claude

PR #325, opened 2026-08-21
OFF-369: Create compatibility spec for HTTP Schema

Kalin-Claude:
Adds `oxygen.http.schema.compat` — the HTTP-layer analogue of `oxygen.schema.compat`, i.e. Phase 4 ("API-incompatibility detection", spec/comparison half) of the HTTP feature plan.

## What
- `HttpCompat.compare(from, to): HttpComparisonResult` — normalizes both specs (`withoutLineNos`), then diffs apis → endpoints by name, and per endpoint: method, path shapes, query/header params, request & response bodies.
- Every **type-reference** transition delegates to the type-level `oxygen.schema.compat.Compared.compareRoot` and its `ComparisonResult` is interpreted **directionally by variance**: request inputs (path/query/header/request-body) are contravariant (compatible ⇔ accepts a superset); responses (response-body/response-header) are covariant (compatible ⇔ produces a subset). Both specs resolve against one merged schema bundle (the type-level `Compared` resolves both refs in a single bundle).
- `HttpComparisonResult` — an accumulating monoid of `HttpBreak` (breaking) + `HttpChange` (backwards-compatible) findings, each tagged with an `HttpLocation`; `isCompatible` + a three-way `HttpCompatibility` summary (`ExactEqual` / `BackwardsCompatible` / `Breaking`). No `JsonCodec` (an in-memory result, like `ComparisonResult`).

## Scope
Spec-only. Persisting a spec + failing CI on a breaking result stays **OXY-38**; this library is pure and I/O-free.

The exact compatibility table was an open question on the ticket, so the implemented table is a documented default grounded in the ticket's stated examples (`agent-docs/http-docs/http-feature-plan.md`, Phase 4) — not a hard spec. Status codes / `caseStatuses` / `mcp` are treated as descriptive (not compared), matching the compiled model's own "de-emphasized" stance.

## Tests
`modules/http/it-test/.../HttpCompatSpec.scala` — 10 tests, green: remove endpoint = breaking, add endpoint = compatible, method change = breaking, add required param = breaking / add optional = compatible, narrow request body = breaking, widen sum response = compatible, self-compare = exact-equal, JSON round-trip, line-number stability.

Verified: `oxygen-httpJVM/compile`, `http-it/testOnly oxygen.http.HttpCompatSpec` (10 passed), `scalafmtCheck` all green.

Jira: OFF-369

### DIFF
--- a/modules/http/it-test/src/test/scala/oxygen/http/HttpCompatSpec.scala
+++ b/modules/http/it-test/src/test/scala/oxygen/http/HttpCompatSpec.scala
@@ -1,0 +1,120 @@
+package oxygen.http
+
+import oxygen.http.schema.compat.*
+import oxygen.http.schema.compiled.*
+import oxygen.http.server.DeriveEndpoints
+import oxygen.json.JsonCodec
+import oxygen.predef.core.*
+import oxygen.predef.test.*
+import oxygen.schema.compiled.CompiledSchemaRef
+import scala.annotation.experimental
+import scala.collection.immutable.ArraySeq as IArraySeq
+import zio.http.Method
+
+/**
+  * Tests for `oxygen.http.schema.compat.HttpCompat` — the HTTP-schema compatibility diff (OFF-369).
+  *
+  * Fixtures are built from a real compiled [[RawCompiledApiSpec]] ([[CompatFixtureApi]]); the `to` side
+  * of each case is a structural mutation of the base endpoints over the '''shared''' schema bundle, so
+  * every rule from the ticket is asserted directly.
+  */
+@experimental
+object HttpCompatSpec extends OxygenSpecDefault {
+
+  private val base: RawCompiledApiSpec =
+    CompiledApiSpec.compileWithoutLineNos(summon[DeriveEndpoints[CompatFixtureApi]].endpoints.toArraySeq.map(_.schema))
+
+  private val endpoints: ArraySeq[RawCompiledEndpoint] = base.apis.flatMap(_.endpoints)
+  private val small: RawCompiledEndpoint = endpoints.find(_.name == "takesSmall").get
+  private val big: RawCompiledEndpoint = endpoints.find(_.name == "takesBig").get
+
+  private val smallBodyRef: CompiledSchemaRef =
+    small.request.body match {
+      case RawCompiledRequestBody.Single(_, _, ref) => ref
+      case RawCompiledRequestBody.Empty             => sys.error("fixture defect: takesSmall has no request body")
+    }
+
+  /** Build a single-api spec from the given endpoints, reusing the full shared bundle. */
+  private def spec(es: RawCompiledEndpoint*): RawCompiledApiSpec =
+    RawCompiledApiSpec(IArraySeq(RawCompiledApi(None, None, IArraySeq.from(es))), base.schemas)
+
+  private def withMethod(ep: RawCompiledEndpoint, method: Method): RawCompiledEndpoint =
+    ep.copy(request = ep.request.copy(method = Some(method)))
+
+  private def addQueryParam(ep: RawCompiledEndpoint, name: String, kind: CompiledParamType): RawCompiledEndpoint =
+    ep.copy(request = ep.request.copy(queryParams = ep.request.queryParams :+ RawCompiledParam(name, None, kind, smallBodyRef)))
+
+  private def withRequestBody(ep: RawCompiledEndpoint, body: RawCompiledRequestBody): RawCompiledEndpoint =
+    ep.copy(request = ep.request.copy(body = body))
+
+  private def withSuccessBody(ep: RawCompiledEndpoint, body: RawCompiledResponseBody): RawCompiledEndpoint =
+    ep.copy(successResponse = ep.successResponse.copy(body = body))
+
+  override def testSpec: TestSpec =
+    suite("HttpCompatSpec")(
+      test("a spec compared to itself is exact-equal") {
+        val res = HttpCompat.compare(spec(small, big), spec(small, big))
+        assertTrue(res.isCompatible, res.compatibility == HttpCompatibility.ExactEqual, res.breaks.isEmpty, res.changes.isEmpty)
+      },
+      test("removing an endpoint is breaking") {
+        val res = HttpCompat.compare(spec(small, big), spec(small))
+        assertTrue(
+          !res.isCompatible,
+          res.compatibility == HttpCompatibility.Breaking,
+          res.breaks.exists { case b: HttpBreak.EndpointRemoved => b.name == "takesBig"; case _ => false },
+        )
+      },
+      test("adding an endpoint is backwards-compatible") {
+        val res = HttpCompat.compare(spec(small), spec(small, big))
+        assertTrue(res.isCompatible, res.compatibility == HttpCompatibility.BackwardsCompatible)
+      },
+      test("changing the HTTP method is breaking") {
+        val res = HttpCompat.compare(spec(small), spec(withMethod(small, Method.GET)))
+        assertTrue(
+          !res.isCompatible,
+          res.breaks.exists { case _: HttpBreak.MethodChanged => true; case _ => false },
+        )
+      },
+      test("adding an optional query param is backwards-compatible") {
+        val res = HttpCompat.compare(spec(small), spec(addQueryParam(small, "extra", CompiledParamType.Optional)))
+        assertTrue(res.isCompatible, res.compatibility == HttpCompatibility.BackwardsCompatible)
+      },
+      test("adding a required query param is breaking") {
+        val res = HttpCompat.compare(spec(small), spec(addQueryParam(small, "extra", CompiledParamType.Required)))
+        assertTrue(
+          !res.isCompatible,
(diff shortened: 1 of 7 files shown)


modules/http/zio/src/main/scala/oxygen/http/schema/compat/HttpCompat.scala
```diff
@@ -0,0 +1,291 @@
...
+    */
+  def compare(from: RawCompiledApiSpec, to: RawCompiledApiSpec): HttpComparisonResult = {
+    val f: RawCompiledApiSpec = from.withoutLineNos
+    val t: RawCompiledApiSpec = to.withoutLineNos
+    // The type-level `Compared` resolves BOTH refs against a single bundle (see `Comparison.comparePostCheck`,
+    // and `CompatSpec` which evals with `full, full`). So diff the two specs against ONE merged bundle
+    // (deduped by ref); every `from`/`to` type ref must be resolvable there. Same-name-but-changed types
+    // collide by ref — a known limit of the shared machinery; version-distinct type names diff exactly.
+    val schemas: FullCompiledSchemas = mergedSchemas(f, t)
+    compareSpec(f, t).eval(schemas, schemas).result
```
Kalin-Claude, line 51: **major** — this is the load-bearing correctness gap. `mergedSchemas` (below) dedupes `from ++ to` `distinctBy(_.ref)`, and `eval(schemas, schemas)` hands the type-level machinery one bundle for both sides. `Comparison.comparePostCheck` resolves *both* refs against the first bundle (`fromSchemas.resolve(toRef)`), so any type whose ref/name is unchanged between versions resolves to a single entry and compares `ExactEqual`.

Concrete failure: persist v1 with `final case class User(id: String)`, then in v2 add `email: String` (required). Ref stays `User`; `compare` returns `BackwardsCompatible`/`ExactEqual` — no `IncompatibleType` break — even though old clients now fail to satisfy a required field. This is precisely the OXY-38 CI case (stable type names across versions), and it's the single most common breaking change a compat gate exists to catch. Every test dodges it by *renaming* the type (`Small`/`Big`, `Shape`/`ShapeWide`), so the gap is untested.

The doc calls it a "known limit," but as written the tool is fail-*open* on in-place type edits. PR #326 walks each side's own bundle specifically to avoid this. Please reconcile the two designs — either resolve `to` against `toSchemas` (needs the type-level `comparePostCheck` fix) or namespace refs by version before the merge — and add a test where a same-named type gains a required field.

---
:robot: claude

Kalin-Claude (commented):
Adversarial standards + correctness review (reviewer role, fresh pass — no prior review threads).

**Verdict: COMMENT / changes requested.** The variance model is sound and I verified it end-to-end (see below). But the single-bundle resolution design has a load-bearing correctness gap for the tool's central purpose, and this package collides head-on with PR #326.

### Blocking

1. **major — the single-bundle resolution cannot detect the most common breaking change: an existing type modified in place.** `compare` merges both specs' schema bundles `distinctBy(_.ref)` and `eval`s with `(schemas, schemas)`; the type-level `Comparison.comparePostCheck` resolves *both* `fromRef` and `toRef` against the *first* bundle (`fromSchemas.resolve(toRef)`), so a type that keeps its name across versions (`User` → `User` with an added required field) collapses to one entry and diffs `ExactEqual` — a false `compatible`. In the real OXY-38 CI scenario (persist old spec, compare to freshly-compiled current code) type names are stable across versions, so in-place field additions/removals — the bread-and-butter breaking change a compat gate exists to catch — are silently invisible. The 10 tests all sidestep this by using *renamed* types (`Small`/`Big`, `Shape`/`ShapeWide`). It's documented as a "known limit," but for a compatibility checker this undersells it to the point of being fail-open. Inline at the merge/eval site.

### Cross-cutting: collision with PR #326 (OFF-352)

**major (architectural) — #325 and #326 both introduce `oxygen.http.schema.compat` independently, and will conflict.** Overlap I can see from the two diffs:
- Both add the *same package* with divergent public APIs: this PR's `HttpCompat.compare → HttpComparisonResult` (+ `HttpBreak`/`HttpChange`/`HttpVariance`/`HttpLocation`/`HttpCompatibility`) vs #326's `HttpApiSpecComparison.compare → Result` (+ `BreakingChange`, `Compatibility`, `Dir`, and a `.jvm` `HttpApiSpecCheck`). Two parallel implementations of Phase 4, same package — whichever merges second leaves a duplicated/divergent surface.
- Both edit the **same** `agent-docs/http-docs/http-feature-plan.md` Phase 4 section — a guaranteed textual merge conflict.
- Both add overlapping it-test fixtures (`CompatFixtureApi` here vs `ApiSpecCompatFixtures` there).

Notably, #326's own module doc **explicitly rejects this PR's approach for exactly finding #1**: it walks the structure directly against each side's *own* bundle *"rather than delegating to `oxygen.schema.compat` — whose comparison assumes both refs live in a single bundle and so cannot compare two independently-compiled specs where a type keeps its name but changes in place (exactly the common case here)."* These two PRs need to be reconciled into one design before either merges; the divergence on the single-bundle question is the crux, not a cosmetic conflict.

### Non-blocking

- **minor** — over-conservative false-positive breaks (fail-closed, so safe but noisy): `pathShape` embeds the path-param *name* (`{$name}`), so renaming a path param (wire-positional, wire-compatible) reports `PathRemoved` = breaking; and `enumBreaking` treats any `caseSensitive`/`exhaustive` difference as breaking regardless of variance. Inline.
- **nit** — test-fixture partial extraction / throw: `endpoints.find(...).get` (×2) and `sys.error(...)` in `HttpCompatSpec` object init. Test-only and infra-ish, but `.get`/`throw` are on the naughty list; `fail(...)`/pattern-match reads cleaner. Inline.

### Cleared suspicions (checked, not flagged)

- **Variance direction is correct.** `FromIsMoreSpecific` (`to` wider) breaks only Covariant/outputs; `ToIsMoreSpecific` (`to` narrower) breaks only Contravariant/inputs; product added-required-field breaks inputs, removed-required-field breaks outputs; `requirednessBreak` flips correctly (input →required, output →optional). All consistent with the documented table.
- **`s"…"` in the `message`/`toIndentedString` rendering is *not* a `str"…"` violation** — the mirrored type-level `oxygen.schema.compat.ComparisonResult` (which this is explicitly "the analogue of") renders with plain `s"…"` + `IndentedString`, not `Text`/`str`. Flagging it would be inconsistent with the established local convention.
- **Braces** — the un-braced `match`es (`isRequired`, `responseBodyKind`, `requirednessBreak`) are all trivial one-line-RHS matches, within the allowed exception.

---
:robot: claude

modules/http/zio/src/main/scala/oxygen/http/schema/compat/HttpCompat.scala
```diff
@@ -0,0 +1,291 @@
...
+      headers <- compareParams(loc / "header", from.headers, to.headers, HttpVariance.Contravariant)
+      body <- compareRequestBody(loc / "body", from.body, to.body)
+    } yield method ++ paths ++ query ++ headers ++ body
+  }
+
+  private def pathShape(p: RawCompiledPaths): String = {
+    val segs: String =
+      p.segments.iterator.map {
+        case RawCompiledPathSegment.Const(path)       => path
+        case RawCompiledPathSegment.Param(name, _, _) => s"{$name}"
```
Kalin-Claude, line 109: **minor** — `pathShape` embeds the path-param *name* into the shape string (`{$name}`). Path params are wire-positional, so renaming `{id}` → `{userId}` is wire-compatible, yet the shape set diff reports it as `PathRemoved` (breaking) + a path-added change. Fail-closed (safe, but a spurious hard-break on a rename). Consider shaping path params positionally (e.g. `{}`) and diffing param *types* by position rather than by name.

(Same over-conservative flavor at `enumBreaking:273`: `caseSensitive`/`exhaustive` differences are treated as breaking in *both* variances — a covariant/output-only relaxation would still trip a break.)

---
:robot: claude

modules/http/it-test/src/test/scala/oxygen/http/HttpCompatSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+  */
+@experimental
+object HttpCompatSpec extends OxygenSpecDefault {
+
+  private val base: RawCompiledApiSpec =
+    CompiledApiSpec.compileWithoutLineNos(summon[DeriveEndpoints[CompatFixtureApi]].endpoints.toArraySeq.map(_.schema))
+
+  private val endpoints: ArraySeq[RawCompiledEndpoint] = base.apis.flatMap(_.endpoints)
+  private val small: RawCompiledEndpoint = endpoints.find(_.name == "takesSmall").get
+  private val big: RawCompiledEndpoint = endpoints.find(_.name == "takesBig").get
```
Kalin-Claude, line 29: **nit** — `.get` on the `find` `Option` (here and line 28), plus `sys.error(...)` in the body-ref extractor (line 34), are partial-extraction / `throw` on the naughty list. Test-only object init so low-stakes, but `.getOrElse(fail(...))` / a pattern-match with `fail` keeps it off the list and gives a better failure than a raw NoSuchElement/RuntimeException if the fixture drifts.

---
:robot: claude

PR #327, opened 2026-08-31
Add JWTHeaderAndPayload (crypto-model) + TypedJWT (schema) to oxygen

Kalin-Claude:
## What

Two constructs for minting / wrapping JWTs with a **custom, typed** claim payload — the upstream home for concepts previously sketched/vendored downstream.

### `oxygen-crypto-model` — `JWTHeaderAndPayload[A]`
A first-class, signable base64url header+payload (the exact JWS signing input), so a caller can mint a JWT with a **custom** payload and control the signature step, rather than accept oxygen's opinionated `JWT.StandardPayload` issuance.

```
JWTHeaderAndPayload.calculate(header, payload)
  .`headerBase64.payloadBase64`                  // JWS signing input
  -> SignatureService.Signer.signBase64(input)   // -> Signature.Base64
  -> .withSignature(sig)                          // -> JWT[A]
```

### `oxygen-schema` — `TypedJWT[A]`
A base for a domain **typed JWT** (a `JWT[A]` wrapper), following the `CurrencyClass` idiom (`type X = X.Type; object X extends TypedJWT[Claims]`). Requires a `JsonSchema[A]`.

- **`Type`** — the ergonomic token, plus the one HTTP-optimized `PlainTextSchema` (via `PlainTextSchema.jwt`, standard `Authorization: Bearer …`); use this when the header format is well-documented. The four variants are plain `PlainTextSchema.string` transforms by contrast. Also carries programmatic decoders and the `HeaderAndPayload` sign→assemble flow.
- **Four granular `<: Type = Type` wire variants** — for pinning an exact decode/encode; each has `apply`/`unwrap`, a `given Conversion[Type, _]`, and a `PlainTextSchema` (`PlainTextSchema.string.transformOrFail` over the `JWT.decode*` fns):
  - `Bearer` — decode+encode strict `Bearer …`
  - `Token` — decode+encode strict bare token
  - `BearerOrToken` / `TokenOrBearer` — both decode either wire form; each **encodes the form named first** (`BearerOrToken` → `Bearer …`, `TokenOrBearer` → bare token)

## Notes

- `JWTHeaderAndPayload` builds on `BearerToken`'s `private[crypto]` ctor legitimately, in-package.
- ⚠️ **Intentional layering wrinkle (signed off):** `TypedJWT` lives in `oxygen-schema` (which depends on `oxygen-crypto-model`, not the reverse) because it needs `PlainTextSchema`/`JsonSchema` — but keeps package `oxygen.crypto.model` so the typed-JWT machinery reads as one namespace with `JWT`/`JWTHeaderAndPayload`.
- 2 new files (+168 / −0); compiles green (`oxygen-crypto-modelJVM`, `oxygen-schemaJVM`).

https://claude.ai/code/session_016wZn74phFB3Jn1npzWo2co

### DIFF
--- a/modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
+++ b/modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
@@ -1,0 +1,121 @@
+package oxygen.crypto.model
+
+import oxygen.core.ConversionUtils
+import oxygen.json.{JsonDecoder, JsonEncoder}
+import oxygen.schema.{JsonSchema, PlainTextSchema}
+
+/**
+  * Base for a **typed JWT**: a domain wrapper around a `JWT[A]` whose payload is a specific claim type `A`.
+  *
+  * Concrete tokens follow the same idiom as `CurrencyClass` — an `object` that `extends TypedJWT[Claims]`,
+  * plus a top-level alias exposing the opaque member:
+  * {{{
+  *   type MyToken = MyToken.Type
+  *   object MyToken extends TypedJWT[MyClaims]
+  * }}}
+  * so `MyToken` names the opaque token type and `MyToken.<…>` constructs / decodes it.
+  *
+  * Requires a `JsonSchema[A]` for the payload (from which both the encoder used to sign and the decoder used
+  * to validate are drawn).
+  *
+  * NOTE: this lives in `oxygen-schema` (not `oxygen-crypto-model`) because it needs `PlainTextSchema` /
+  * `JsonSchema`, and `oxygen-schema` depends on `oxygen-crypto-model` (not the reverse). The package is kept
+  * as `oxygen.crypto.model` deliberately, so the typed-JWT machinery reads as one namespace with `JWT` /
+  * `JWTHeaderAndPayload` even though it is compiled in the schema module.
+  */
+abstract class TypedJWT[A](using schema: JsonSchema[A]) {
+
+  private given jsonEncoder: JsonEncoder[A] = schema.jsonEncoder
+  private given jsonDecoder: JsonDecoder[A] = schema.jsonDecoder
+
+  /**
+    * The typed JWT itself (a `JWT[A]`) — the ergonomic "work with it" type, and the one to reach for when the
+    * HTTP auth header format is already well-documented / fixed. When you need to pin an *exact* wire
+    * encode/decode, use the granular [[Bearer]] / [[Token]] / [[BearerOrToken]] / [[TokenOrBearer]] variants.
+    * Exposed by the extender as `type X = X.Type`.
+    */
+  final opaque type Type = JWT[A]
+  object Type {
+
+    def apply(jwt: JWT[A]): Type = jwt
+
+    extension (self: Type) {
+
+      /** The raw underlying `JWT[A]`. */
+      def toJWT: JWT[A] = self
+
+      /** The decoded claim payload. */
+      def jwtPayload: A = self.payload
+
+      /** The signed [[BearerToken]] — call `.bearer` for `Authorization: Bearer …`, or `.token` for the bare form. */
+      def jwtBearer: BearerToken = self.token
+    }
+
+    def fromBearerToken(token: BearerToken): Either[String, Type] = JWT.decode[A](token).map(apply)
+    def decodeBearer(value: String): Either[String, Type] = JWT.decodeBearer[A](value).map(apply)
+    def decodeToken(value: String): Either[String, Type] = JWT.decodeToken[A](value).map(apply)
+    def decodeBearerOrToken(value: String): Either[String, Type] = JWT.decodeBearerOrToken[A](value).map(apply)
+
+    /** Assemble a token from its signed [[HeaderAndPayload]] (signature from `Signer.signBase64`). */
+    def headerPayloadSignature(hap: HeaderAndPayload, signature: Signature.Base64): Type =
+      apply(hap.withSignature(signature)) // HeaderAndPayload <: JWTHeaderAndPayload[A], so `withSignature` is in reach
+
+    /**
+      * The one HTTP-optimized schema — leverages `PlainTextSchema.jwt` (standard `Authorization: Bearer …`).
+      * The granular variants below are, by contrast, plain `PlainTextSchema.string` transforms.
+      */
+    given schema: PlainTextSchema[Type] = PlainTextSchema.jwt[A].transform(apply, _.toJWT)
+
+  }
+
+  /** The base64url header+payload for THIS token type, ready to sign via `Signer.signBase64`. */
+  final opaque type HeaderAndPayload <: JWTHeaderAndPayload[A] = JWTHeaderAndPayload[A]
+  object HeaderAndPayload {
+    def calculate(header: JWTHeader, payload: A): HeaderAndPayload = JWTHeaderAndPayload.calculate(header, payload)
+  }
+
+  //////////////////////////////////////////////////////////////////////////////////////////////////////
+  //      Granular wire variants — pin an exact decode/encode for schema-driven (de)serialization.
+  //      Each is `<: Type = Type` (so it IS a `Type`), with `decode` + an identity `Conversion[Type, _]`.
+  //////////////////////////////////////////////////////////////////////////////////////////////////////
+
+  /** Decodes AND encodes the strict `Bearer <header>.<payload>.<signature>` form. */
+  final opaque type Bearer <: Type = Type
+  object Bearer {
+    def apply(value: Type): Bearer = value
+    def decode(value: String): Either[String, Bearer] = Type.decodeBearer(value)
+    given Conversion[Type, Bearer] = ConversionUtils.id
+    given schema: PlainTextSchema[Bearer] = PlainTextSchema.string.transformOrFail(decode, Type.jwtBearer(_).bearer)
+  }
+
+  /** Decodes AND encodes the strict bare `<header>.<payload>.<signature>` token form (no `Bearer ` prefix). */
+  final opaque type Token <: Type = Type
+  object Token {
+    def apply(value: Type): Token = value
+    def decode(value: String): Either[String, Token] = Type.decodeToken(value)
+    given Conversion[Type, Token] = ConversionUtils.id
+    given schema: PlainTextSchema[Token] = PlainTextSchema.string.transformOrFail(decode, Type.jwtBearer(_).token)
+  }
+
+  // Naming convention for the "decode either" pair: BOTH decode either wire form; each ENCODES to the form
+  // named FIRST — `BearerOrToken` encodes `Bearer …`, `TokenOrBearer` encodes the bare token.
+
+  /** Decodes EITHER wire form; encodes the FIRST-named form → `Bearer <header>.<payload>.<signature>`. */
+  final opaque type BearerOrToken <: Type = Type
+  object BearerOrToken {
+    def apply(value: Type): BearerOrToken = value
+    def decode(value: String): Either[String, BearerOrToken] = Type.decodeBearerOrToken(value)
+    given Conversion[Type, BearerOrToken] = ConversionUtils.id
+    given schema: PlainTextSchema[BearerOrToken] = PlainTextSchema.string.transformOrFail(decode, Type.jwtBearer(_).bearer)
+  }
+
+  /** Decodes EITHER wire form; encodes the FIRST-named form → the bare `<header>.<payload>.<signature>` token. */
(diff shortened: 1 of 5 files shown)


modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  /** Decodes AND encodes the strict `Bearer <header>.<payload>.<signature>` form. */
+  final opaque type Bearer <: Type = Type
+  object Bearer {
+    def apply(value: Type): Bearer = value
+    extension (self: Bearer) def unwrap: Type = self
+    given Conversion[Type, Bearer] = apply(_)
+    given schema: PlainTextSchema[Bearer] =
+      PlainTextSchema.string.transformOrFail(JWT.decodeBearer[A](_).map(jwt => apply(Type(jwt))), _.token.bearer)
+  }
```
Kalin-Rudnicki, line 89: in each of the 4 companion
1. `def unwrapType`
2. `schema` calls this
3. all the wrapping is pointless - `JWT.decode___` returns `JWT[A]`, which within the companion, should type correctly? If theres a 2-level-deep problem, then calling the `Type.decode___` function should be typed correctly...

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
```diff
@@ -0,0 +1,112 @@
...
+  private given jsonDecoder: JsonDecoder[A] = schema.jsonDecoder
+
+  /**
+    * The typed JWT itself (a `JWT[A]`) — the ergonomic "work with it" type, and the one to reach for when the
+    * HTTP auth header format is already well-documented / fixed. When you need to pin an *exact* wire
+    * encode/decode, use the granular [[Bearer]] / [[Token]] / [[BearerOrToken]] / [[TokenOrBearer]] variants.
+    * Exposed by the extender as `type X = X.Type`.
+    */
+  final opaque type Type <: JWT[A] = JWT[A]
+  object Type {
```
Kalin-Rudnicki: 1. remove the `<: JWT[A]`
2. add an `unwrapJWT` extension

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Done in 7974f3ac (all 4 variant companions):
1. renamed the extension `unwrap` → `unwrapType`;
2. `schema` now uses it — encode is `_.unwrapType.unwrapJWT.token.<bearer|token>`;
3. dropped the pointless wrapping — decode goes through `Type.decode…` (already lifts `JWT[A]` → `Type`), so it's just `Type.decodeBearer(_).map(apply)`.

(`unwrapJWT` is Type's extension from the other thread; imported once at class level so the variants can reach it.)

---
:robot: claude

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Done in 7974f3ac:
1. removed the `<: JWT[A]` bound → `final opaque type Type = JWT[A]`;
2. added `extension (self: Type) def unwrapJWT: JWT[A]`.
`Type`'s schema now encodes via `_.unwrapJWT`, and the granular variants reach the raw JWT through `unwrapType.unwrapJWT`.

---
:robot: claude

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  }
+
+  /** The base64url header+payload for THIS token type, ready to sign via `Signer.signBase64`. */
+  final opaque type HeaderAndPayload <: JWTHeaderAndPayload[A] = JWTHeaderAndPayload[A]
+  object HeaderAndPayload {
+    def calculate(header: JWTHeader, payload: A): HeaderAndPayload = JWTHeaderAndPayload.calculate(header, payload)
+  }
+
+  import Type.unwrapJWT // make the extension available to the granular variants' encoders below
```
Kalin-Rudnicki: this should not be necessary

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
```diff
@@ -0,0 +1,115 @@
...
+    * The typed JWT itself (a `JWT[A]`) — the ergonomic "work with it" type, and the one to reach for when the
+    * HTTP auth header format is already well-documented / fixed. When you need to pin an *exact* wire
+    * encode/decode, use the granular [[Bearer]] / [[Token]] / [[BearerOrToken]] / [[TokenOrBearer]] variants.
+    * Exposed by the extender as `type X = X.Type`.
+    */
+  final opaque type Type = JWT[A]
+  object Type {
+
+    def apply(jwt: JWT[A]): Type = jwt
+    extension (self: Type) def unwrapJWT: JWT[A] = self
```
Kalin-Rudnicki: actually, rename this, and have 3 functions:
1. `def toJWT`
2. `jwtPayload`
3. `jwtBearer`

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  /** Decodes AND encodes the strict bare `<header>.<payload>.<signature>` token form (no `Bearer ` prefix). */
+  final opaque type Token <: Type = Type
+  object Token {
+    def apply(value: Type): Token = value
+    extension (self: Token) def unwrapType: Type = self
+    given Conversion[Type, Token] = apply(_)
+    given schema: PlainTextSchema[Token] =
+      PlainTextSchema.string.transformOrFail(Type.decodeToken(_).map(apply), _.unwrapType.unwrapJWT.token.token)
+  }
```
Kalin-Rudnicki, line 98: add a `def decode(value: String): Either[String, ...]` in each of the `object`

the `transformOrFail` should call it

once the 3 helpers are in `Type` companion, Im pretty sure all you should need is:
```scala
PlainTextSchema.string.transformOrFail(decide, _.jwtBearer)
```

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Done in a20c9d17. Replaced `unwrapJWT` with helpers on the `Type` companion: `toJWT`, `jwtPayload`, `jwtBearer` — plus `jwtToken` (a 4th, needed to encode the bare-token variants). `Type`'s schema encodes via `toJWT`.

---
:robot: claude

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Done in a20c9d17. Each variant now has `def decode(value: String): Either[String, …]`, and the schema is `PlainTextSchema.string.transformOrFail(decode, Type.jwtBearer(_))` (or `jwtToken` for the token-encoders). Dropped the now-redundant `unwrapType`.

---
:robot: claude

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Removed the import — sidestepped it by calling the helpers in **prefix** form, `Type.jwtBearer(_)`, in the variant schemas.

Heads-up though: postfix `_.jwtBearer` genuinely *does* require `import Type.jwtBearer`. The helpers live in `object Type`, and a **path-dependent opaque type's companion extensions aren't in a sibling object's implicit scope** — the compiler itself suggests the import. Prefix form avoids it, so there's no import now.

---
:robot: claude

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Update: the `unwrapType` extension this thread asked for is **gone** as of a20c9d17 — superseded by the `Type`-companion helpers (`toJWT`/`jwtPayload`/`jwtBearer`/`jwtToken`) from the later threads, which the variant schemas now use directly. Flagging since this thread's earlier "done" no longer holds — say the word if you want `unwrapType` back as a standalone accessor.

---
:robot: claude

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
```diff
@@ -0,0 +1,119 @@
...
+    given Conversion[Type, Bearer] = apply(_)
+    given schema: PlainTextSchema[Bearer] = PlainTextSchema.string.transformOrFail(decode, Type.jwtBearer(_))
+  }
+
+  /** Decodes AND encodes the strict bare `<header>.<payload>.<signature>` token form (no `Bearer ` prefix). */
+  final opaque type Token <: Type = Type
+  object Token {
+    def apply(value: Type): Token = value
+    def decode(value: String): Either[String, Token] = Type.decodeToken(value).map(apply)
+    given Conversion[Type, Token] = apply(_)
```
Kalin-Rudnicki: there should be some oxygen identity conversion you can use here

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
```diff
@@ -0,0 +1,119 @@
...
+    def decode(value: String): Either[String, Bearer] = Type.decodeBearer(value).map(apply)
+    given Conversion[Type, Bearer] = apply(_)
+    given schema: PlainTextSchema[Bearer] = PlainTextSchema.string.transformOrFail(decode, Type.jwtBearer(_))
+  }
+
+  /** Decodes AND encodes the strict bare `<header>.<payload>.<signature>` token form (no `Bearer ` prefix). */
+  final opaque type Token <: Type = Type
+  object Token {
+    def apply(value: Type): Token = value
+    def decode(value: String): Either[String, Token] = Type.decodeToken(value).map(apply)
```
Kalin-Rudnicki: are you 1000% sure you need that `.map(apply)`, Im almost certain its unnecessary

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
```diff
@@ -0,0 +1,119 @@
...
+
+    extension (self: Type) {
+      /** The raw underlying `JWT[A]`. */
+      def toJWT: JWT[A] = self
+      /** The decoded claim payload. */
+      def jwtPayload: A = self.payload
+      /** Wire string in `Authorization: Bearer …` form. */
+      def jwtBearer: String = self.token.bearer
+      /** Wire string in bare `<header>.<payload>.<signature>` token form. */
+      def jwtToken: String = self.token.token
```
Kalin-Rudnicki: `def jwtBearer: BearerToken`, and then the caller can call whichever encode function they want, drop the other one

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Done in a24c6b0a — `given Conversion[Type, _] = ConversionUtils.id` in all 4 (oxygen's identity conversion — same pattern as `Conversion[Text, Text.Auto] = ConversionUtils.id`).

---
:robot: claude

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Done in a24c6b0a — you're right, dropped `.map(apply)`. `Type.decode…` returns `Either[String, Type]`, and inside the variant companion `<Variant>` is transparent (`= Type`), so it already satisfies `Either[String, <Variant>]` directly.

---
:robot: claude

modules/general/schema/src/main/scala/oxygen/crypto/model/TypedJWT.scala
Kalin-Claude: Done in a24c6b0a — `jwtBearer` now returns the `BearerToken`; dropped `jwtToken`. The variant schemas encode via `Type.jwtBearer(_).bearer` (bearer-encoders) / `Type.jwtBearer(_).token` (token-encoders).

---
:robot: claude

